PR #191, opened 2023-11-28
Support array types

plcplc:
### What

We now recognize array types in several places:

* Database introspection columns and comparison operators of array type
* Columns of array type can be queried
* Arguments of array type can be input
* Arrays work in tables and native queries alike

Notable limitations, due in part to the modest expressiveness of the ndc-spec in its current state:

* There is no way to filter array fields
* There is no way to project fields from the elements of an array field
* We only support monomorphic comparison operators

### How

The introspection query has had its `types` sub-query split into `scalar_types` and `array_types` subqueries. We benefit from the fact that postgres only supports one-dimensional array _types_ (i.e., array dimensionality is a dynamic aspect of an array value), so we don't have to recurse to construct the json-representation of array types.

Most of the rust-language changes of this PR follow mechanically from changing `metadata::ScalarType` to the new `metadata::Type`, which also models array types.

However, since our field types now include arrays we need to observe that certain parts (e.g. ComparisonOperators) only accept base scalar types (e.g. in `version1.rs`, `filtering.rs`).

In defining the new type for column types we also take care to maintain backwards compatibility, which means deserialization gets a bit more verbose than just the automatically derived implementations we used to get by with (see `metadata/database.rs`).

Since we also support array-typed arguments we have to be able to output SQL array constructors (see `values.rs`, `sql/convert.rs`).

We introduce two new tests:

* One test which just verifies that we can output an array column from a native query
* Another which verifies that we can take an array-valued argument to a native query.

### Update

Altering the representation of column types in the configuration constitues a breaking change. Even if we are still able to ingest the historic configuration formats these are not described by the published json schema, and we would only ever output in the newest format.

In order to address the issue of versioning the original contents of this PR is being split in two:
* The first part (this PR) contains the logic for dealing with array types, and includes a new level of indirection between the api transport types and those used internally.
* A second follow-up PR will introduce a version 2 of the configuration format, update introspection to include array types, and add tests of the same.

An interesting consequence of introducing the distinction between api transport types and internal business types is that the various `tables.json` files of the query-engine translation tests contain json-serialized data of _business_ types.

Worth highlighting is this means we cannot simply copy a subsection of a deployment configuration into a `tables.json` file like we used to, as the business types do not necessarily correspond to any api-version.

While this may seem confusing it also has some benefits:
* Test files have no version drift. They are always representative of the current internal business data model
* The test remains coupled with the query-engine crate rather than the connector crate.

Updating test files can become a chore too. There is no tool that can upgrade them automatically like we do for deployments, and it would probably not be worth the effort to try and make one.

What I ended up doing in the particular case of this PR was a small shell script:
```bash
for f in $(fd -t f 'tables\.json')
do
  jq 'walk(if type == "object" and has("type") then .type |= {scalarType: .} else . end)' < "$f" \
  | sponge "$f"
done
```

The `jq` script matches any object that has a `type` key and wraps its value with a `{"scalarType": ..}` object. Happily this covered the full breadth of my change.

[sponge](https://joeyh.name/code/moreutils/) is a tool I just discovered which buffers all the stdin it receives and writes everything into the file it's given. Essentially this endows our command with the ability to make in-place edits with shell pipe filters.

### DIFF
--- a/.buildkite/release-config-server.yaml
+++ b/.buildkite/release-config-server.yaml
@@ -52,6 +52,8 @@ steps:
     depends_on:
       - step: rollback-block
     command: .buildkite/scripts/release-component.sh "ndc-postgres-config-server"
+    env:
+      ROLLBACK: "true"
     plugins:
       - hasura/smooth-secrets#v1.3.2:
           secrets:
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -13,6 +13,8 @@ jobs:
     runs-on: ubuntu-latest
     # only run if the previous workflow succeeded,
     # and only on the `main` branch or version tags
+    # note we currently tag the image with 'latest', so will want to
+    # stop doing so if we run this on PR branches etc
     if: |
       github.event.workflow_run.conclusion == 'success'
         && (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/v'))
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,6 @@ generated/
 
 # deployments created for testing
 static/temp-deploys/*.json
+
+# IntelliJ IDEs
+.idea
--- a/ci/deploy.sh
+++ b/ci/deploy.sh
@@ -49,6 +49,7 @@ function run {
 #
 # Additionally sets a branch tag assuming this is the latest tag for the given
 # branch. The branch tag has the form: dev-main
+# Also sets the 'latest' tag
 # Also sets a tag with just the branch short hash
 function set_dev_tags {
     local branch="$1"
@@ -60,7 +61,7 @@ function set_dev_tags {
     local short_hash
     short_hash="$(git rev-parse --short=9 HEAD)"
     version="${branch_prefix}-${short_hash}"
-    export docker_tags=("$version" "$branch_prefix" "$short_hash")
+    export docker_tags=("$version" "$branch_prefix" "$short_hash" "latest")
 }
 
 # The Github workflow passes a ref of the form refs/heads/<branch name> or
@@ -79,12 +80,12 @@ function set_docker_tags {
     input="$1"
     if [[ $input =~ ^refs/tags/(v.*)$ ]]; then
         local tag="${BASH_REMATCH[1]}"
-        export docker_tags=("$tag")
+        export docker_tags=("$tag" "latest")
     elif [[ $input =~ ^refs/heads/(.*)$ ]]; then
         local branch="${BASH_REMATCH[1]}"
         set_dev_tags "$branch"
     else
-        export docker_tags=()
+        export docker_tags=("latest")
     fi
 }
 
--- a/docs/limitations.md
+++ b/docs/limitations.md
@@ -3,8 +3,6 @@
 ## Query
 
 - Columns as argument fields in native queries are currently not supported.
-- Order by star count aggregates is currently not supported.
-- Order by star count for nested relationships is currently not supported.
 
 ## Mutations
 
--- a/docs/readme.md
+++ b/docs/readme.md
@@ -2,6 +2,7 @@
 
 `ndc-postgres` is a [Hasura](https://hasura.io/) connector for PostgreSQL.
 
+- [Usage](./usage.md)
 - [Architecture](./architecture.md)
 - [Code of Conduct](./code-of-conduct.md)
 - [Debugging](./debugging.md)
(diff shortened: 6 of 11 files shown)


crates/query-engine/metadata/src/metadata/database.rs
```diff
@@ -4,11 +4,55 @@ use schemars::JsonSchema;
...
+            DeserializableType::V1(v1_type) => Type::from(v1_type),
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum DeserializableType {
+    V0(String),
+    V1(V1Type),
```
soupi, line 37: Should this be v1 and v2, to match the configuration version?

soupi: And please add docstrings to the new types :)

plcplc: That is a good question!

I'm not sure if these `V0` and `V1` are/should be related to the `version` field of `Configuration`.

The model of versioning I think I am employing (sort of after-the-fact) is that Configuration.version is bumped only when we make a non-backwards compatible change.

But I'm not at all sure TBH!

plcplc: I think that, after our discussion earlier we should be bumping the Configuration version number as well.

danieljharvey: Probably good to settle on starting at `1`, but versions of individual types like this shouldn't need to go up in lockstep with the configuration version, surely?

crates/query-engine/translation/src/translation/query/values.rs
```diff
@@ -8,39 +8,73 @@ use sql::ast::{Expression, Value};
...
+}
 
-    if should_cast {
-        Ok(sql::ast::Expression::Cast {
-            expression: Box::new(exp),
-            r#type: sql::ast::ScalarType(scalar_type.0.clone()),
-        })
-    } else {
-        Ok(exp)
+pub fn type_to_ast_scalar_type(typ: &database::Type) -> sql::ast::ScalarType {
```
soupi, line 58: docstring please :)

crates/connectors/ndc-postgres/src/configuration/version1.rs
```diff
@@ -455,23 +455,36 @@ pub fn occurring_scalar_types(
...
+            .filter_map(some_scalar_type)
+    });
 
     tables_column_types
         .chain(native_queries_column_types)
         .chain(native_queries_arguments_types)
         .collect::<BTreeSet<metadata::ScalarType>>()
 }
+
+pub fn some_scalar_type(typ: metadata::Type) -> Option<metadata::ScalarType> {
```
soupi, line 485: docstring please!

crates/connectors/ndc-postgres/src/configuration/version1.rs
```diff
@@ -475,3 +475,141 @@ pub fn occurring_scalar_types(
...
+        columns: columns_to_current(&nq.columns),
+        arguments: columns_to_current(&nq.arguments),
+        description: nq.description.clone(),
+    }
+}
+
+/// Metadata information.
+#[derive(Debug, Default, Serialize, Deserialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct Metadata {
```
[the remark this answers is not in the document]
danieljharvey: (or is this the "we need everything nearly identically twice because we're changing one tiny leaf" problem?)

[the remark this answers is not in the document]
plcplc: It _is_ using only the current/newest versions internally :-) 

As much as possible, `version1` continues referring the `metadata` types.

crates/connectors/ndc-postgres/src/configuration.rs
```diff
@@ -21,15 +23,56 @@ pub const CURRENT_VERSION: u32 = 1;
...
+    pub metadata: metadata::Metadata,
 }
 
 impl<'a> version1::Configuration {
     /// Apply the common interpretations on the Configuration API type into an RuntimeConfiguration.
-    pub fn as_runtime_configuration(self: &'a Configuration) -> RuntimeConfiguration<'a> {
+    pub fn as_runtime_configuration(self: &'a Configuration) -> RuntimeConfiguration {
         RuntimeConfiguration {
-            metadata: &self.config.metadata,
+            metadata: metadata_to_current(&self.config.metadata),
```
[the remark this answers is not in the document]
danieljharvey: All I mean is, currently we take the whole `Configuration` and derive `RuntimeConfiguration` from it on each request, surely we could just do this once?

It's not really relevant to this PR, mind, just something I noticed because you made a change near the code that does this.

PR #196, opened 2023-11-30
Comparison functions

i-am-tom:
### What

Currently, we only support comparison operators that happen to be infix operators in Postgres. This PR extends the vocabulary to contain all two-argument functions that return a bool as well. This is particularly useful for geometry code.

### How

A few things going on here.
* We change the introspection code to detect "comparison_procedures" - Postgres functions that operate on two values and return a bool.
* We update the translation code to generate a `FunctionCall` sql ast node in the case it is dealing with a non-infix function.


crates/query-engine/sql/src/sql/ast.rs
```diff
@@ -164,12 +164,6 @@ pub enum Expression {
     },
     /// NOT clause
     Not(Box<Expression>),
-    /// A binary operation on two scalar expression
```
i-am-tom, line 167: Removed in favour of `FunctionCall`

crates/query-engine/sql/src/sql/ast.rs
```diff
@@ -213,10 +207,6 @@ pub enum UnaryOperator {
     IsNull,
 }
 
-/// Represents the name of a binary operator.
```
i-am-tom, line 216: Removed in favour of `FunctionCall`.

crates/query-engine/sql/src/sql/ast.rs
```diff
@@ -225,10 +215,37 @@ pub enum BinaryArrayOperator {
 
 /// A scalar function
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub enum Function {
```
i-am-tom, line 228: We now have all sorts of functions beyond `Coalesce` and `JsonAgg`, and it doesn't seem to buy us much to enumerate all the functions we care about when we're just cramming them into an AST, so `Function` is now a struct.

crates/query-engine/sql/src/sql/convert.rs
```diff
@@ -247,17 +247,6 @@ impl Expression {
                 sql.append_syntax("NOT ");
                 expr.to_sql(sql);
             }
-            Expression::BinaryOperation {
```
i-am-tom, line 250: Absorbed into `FunctionCall`

crates/query-engine/sql/src/sql/convert.rs
```diff
@@ -288,15 +277,25 @@ impl Expression {
...
             }
             Expression::FunctionCall { function, args } => {
-                function.to_sql(sql);
-                sql.append_syntax("(");
-                for (index, arg) in args.iter().enumerate() {
-                    arg.to_sql(sql);
-                    if index < (args.len() - 1) {
-                        sql.append_syntax(", ")
+                if function.is_infix {
+                    sql.append_syntax("(");
```
i-am-tom, line 281: I guess there's a nice follow-up PR to add a `parens` helper that adds a parenthesis, runs a continuation, then adds the closer.

crates/query-engine/sql/src/sql/ast.rs
soupi: Hey, I think it would actually be nice to keep BinaryOperation as well, because essentially we end up have two different implementations in convert.rs with an `if`, so I think it makes more sense to just have two structures. That way we also:
1. mimic the postgres ast as it is
2. we make sure that we build binary operations with exactly two arguments, reducing potential errors with generated queries
3. avoid branching in favor of preserving structures

crates/query-engine/sql/src/sql/ast.rs
```diff
@@ -225,10 +215,41 @@ pub enum BinaryArrayOperator {
...
-    Coalesce,
-    JsonAgg,
-    Unknown(String),
+pub struct Function {
+    pub function_name: String,
+    pub is_infix: bool,
+}
+
+/// The Postgres 'coalesce' function for convenience.
```
soupi, line 223: Cool. I actually think `helpers.rs` is a better place for these, this is where we have similar functions like `true_expr`, etc.

crates/query-engine/sql/src/sql/convert.rs
soupi: If we keep BinaryOperation we wouldn't have to worry about this :)

crates/query-engine/sql/src/sql/ast.rs
i-am-tom: @plcplc

crates/query-engine/sql/src/sql/convert.rs
i-am-tom: Why? Wouldn't this still be a problem for `FunctionCall`?

soupi: Oh sorry. I think I misunderstood your original suggestion.

I'm usually for combinator style in Haskell, but trying to use the same style in Rust I feel like it's more trouble that it's worth because of Rust's imperative nature and type system. I think this current style works well and that

```rs
sql.append_syntax("(");
<thing>
sql.append_syntax(")");
```

isn't really a problem by any means.

static/citus/v2-chinook-deployment.json
```diff
@@ -1683,623 +1683,4484 @@
...
+          "argumentType": "char",
+          "isInfix": true
         },
         "_regex": {
           "operatorName": "~",
-          "argumentType": "char"
+          "argumentType": "char",
+          "isInfix": true
+        },
+        "alter_role_if_exists": {
```
soupi, line 1831: maybe we should filter functions like this one.

plcplc: Yeah that's true. The PR now filters certain types and functions which are to do with server administraton.

crates/tests/databases-tests/src/cockroach/snapshots/databases_tests__cockroach__schema_tests__schema_test__get_schema.snap
```diff
@@ -381,6 +417,18 @@ expression: result
             "type": "named",
             "name": "int2"
           }
+        },
+        "repair_catalog_corruption": {
```
soupi, line 421: I think there are other functions here that should be filtered, `repair_catalog_corruption`, `pg_has_role`, `is_constraint_active`, `set_trace_verbose` are a few names that pop up.

plcplc:
This PR is subsumed by #223

PR #198, opened 2023-12-04
expose procedures as part of the schema endpoint

soupi:
### What

We want to expose the native query procedures we defined in #189 on the `/schema/` endpoint as `procedures` and not `collections`.

### How

- We add a boolean `isProcedure` field to native queries in the `deployment.json` configuration file which specifies whether they are procedures or not (default `false` that is skipped serialization so we maintain backwards compat).
- We include native queries marked is `is_procedure` in the `procedures` list and not the `collections` list.

### DIFF
--- a/crates/connectors/ndc-postgres/src/configuration/version1.rs
+++ b/crates/connectors/ndc-postgres/src/configuration/version1.rs
@@ -615,7 +615,7 @@ pub struct NativeQueryInfo {
     pub arguments: BTreeMap<String, ColumnInfo>,
     #[serde(default)]
     pub description: Option<String>,
-    /// Is this native query a procedure
+    /// True if this native query mutates the database
     #[serde(skip_serializing_if = "std::ops::Not::not")]
     #[serde(default)]
     pub is_procedure: bool,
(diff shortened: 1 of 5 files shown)


crates/connectors/ndc-postgres/src/schema.rs
```diff
@@ -207,9 +208,34 @@ pub async fn get_schema(
     let mut object_types = table_types;
     object_types.extend(native_queries_types);
 
+    let procedures: Vec<models::ProcedureInfo> = metadata
+        .native_queries
+        .0
+        .iter()
+        .filter(|(_, info)| info.is_procedure)
```
soupi, line 215: only include procedures in the procedures list.

plcplc (approved):
Nice!

crates/connectors/ndc-postgres/src/configuration/version1.rs
```diff
@@ -603,13 +604,19 @@ pub struct NativeQueries(pub BTreeMap<String, NativeQueryInfo>);
...
-    /** Columns returned by the Native Query */
+    /// Columns returned by the Native Query
     pub columns: BTreeMap<String, ColumnInfo>,
     #[serde(default)]
-    /** Names and types of arguments that can be passed to this Native Query */
+    /// Names and types of arguments that can be passed to this Native Query
     pub arguments: BTreeMap<String, ColumnInfo>,
     #[serde(default)]
     pub description: Option<String>,
+    /// Is this native query a procedure
```
SamirTalwar, line 618:
```suggestion
    /// True if this native query mutates the database
```

SamirTalwar (approved):

PR #205, opened 2023-12-11
constant folding boolean expressions in exists clauses

soupi:
### What

We didn't cover `EXISTS` clauses in constant folding of boolean expressions. This PR handles them and removes an unnecessary `true AND` case in a test.

### How

normalize the select inside the exists clause.


danieljharvey (approved):

soupi:
@danieljharvey makes sense. There are things where this would work, like this use case perhaps. But it does mean that we'd have to build our own `fold`s and `map`s for example, and enforce that people use the custom functions instead.
My impression is that it is sometimes easier to emit a "messy" ast and clean it up afterwards, and sometimes much harder to clean up afterwards and much easier to just emit clean code. I'm not sure exactly where this particular case fits!

PR #208, opened 2023-12-11
Introduce configuration version 2

plcplc:
### What

This PR adds a new version (`"2"`) of the deployment configuration data format.

This version of the configuration is capable of expressing array types in collections and arguments.

Since this is the first time a new version is introduced there are a lot of changes the only purpose of which is to distinguish between versions.

Only the infrastructure-related shell of the connector is aware of different versions of deployment configurations existing. The core of the connector only works with a single internal version.

This PR is also the one to introduce tests of array types. In hindsight this ought to have been possible in the previous PR that introduced the internal types and transformations (#191).

Note that there is not yet any automated way to upgrade a configuration to a newer version, but this will be introduced shortly.

This PR also adds a changelog entry.

### How

The file `version2.rs` is a duplicate of `version1.rs`, which has been adapted to use the new data types (incidentally these are just the ones of the internal model).

`configuration.sql` now exists as `version1.sql` and `version2.sql` respectively, since these have different capabilities. This is because `version2.sql` introduces the ability to introspect array types.

`configuration.rs` now exposes `RawConfiguration` and `Configuration` types which are enums of all the supported versions (currently 1 and "2").
One big wart on the implementation is that serde and schemars are unable to derive trait implementations for these types correctly, since they only support strings as enum tags, and we used a number literal for version 1.

Once we drop support of version 1 completely we can remove the manually implemented instances.

The various `Connector` trait implementations now explicitly work on the internal representation of a configuration, `RuntimeConfiguration`.


static/aurora/v1-chinook-deployment-template.json
```diff
@@ -687,6 +687,108 @@
         },
         "description": null
       },
+      "delete_playlist_track": {
```
plcplc, line 690: I got errors complaining these native queries were missing from the deployment while I was developing, and so I put that down to a botched merge while I was preparing version 2 of the deployments.

But in this diff I appear to be the one introducing them. Any clarifying info is very welcome.

soupi: I have not yet tested native query mutations for other databases. I was planning on adding more tests but it got delayed because of other stuff.

plcplc: Ah ok. I'm fine to leave these in if you like.

soupi: I'm fine with adding or not adding them in this commit. Whatever is less work for you.

crates/query-engine/metadata/src/metadata/mod.rs
```diff
@@ -11,7 +11,7 @@ use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
 /// Metadata information.
-#[derive(Debug, Default, Serialize, Deserialize, JsonSchema)]
+#[derive(Clone, Debug, Default, Serialize, Deserialize, JsonSchema)]
```
soupi, line 14: I'm wondering where this is used. One nice property we had before was that we only create the metadata at startup and we reuse it.

crates/connectors/ndc-postgres/src/configuration.rs
```diff
@@ -25,54 +77,232 @@ pub const CURRENT_VERSION: u32 = 1;
...
+                res.append(obj);
+                serde_json::value::to_value(res).unwrap()
+            }
+        };
+
+        RawConfigurationCompat(val)
+    }
+}
+
+// Dump of derivied JsonSchema trait implementation, tweaked to put version as an integer for
```
danieljharvey, line 153: We need to remember this is here and needs updating otherwise we're going to have such a confusing time debugging JSON schema.

danieljharvey: If we are able to put the bad generated stuff in a separate file that'd be great for clarity but only if it's easy to do so.

crates/connectors/ndc-postgres/src/configuration/version1.rs
```diff
@@ -451,7 +423,7 @@ fn filter_aggregate_functions(
 
 /// Collect all the types that can occur in the metadata. This is a bit circumstantial. A better
 /// approach is likely to record scalar type names directly in the metadata via configuration.sql.
```
danieljharvey, line 425:
```suggestion
/// approach is likely to record scalar type names directly in the metadata via version1.sql.
```

crates/connectors/ndc-postgres/src/configuration/version2.rs
```diff
@@ -0,0 +1,172 @@
...
+            native_queries: args.metadata.native_queries,
+            aggregate_functions: relevant_aggregate_functions,
+            comparison_operators: relevant_comparison_operators,
+        },
+        configure_options: args.configure_options,
+    })
+}
+
+/// Collect all the types that can occur in the metadata. This is a bit circumstantial. A better
+/// approach is likely to record scalar type names directly in the metadata via configuration.sql.
```
danieljharvey, line 134:
```suggestion
/// approach is likely to record scalar type names directly in the metadata via version2.sql.
```

danieljharvey (approved):

crates/query-engine/metadata/src/metadata/mod.rs
plcplc: Yeah. That property has gone out the window with this change unfortunately, because `as_runtime_configuration` / `RawConfiguration` / `RuntimeConfiguration` no longer track lifetimes like they used to.

This is because I judged that tracking lifetimes would become more complex now that we have `RawConfiguration` being a union type, and with the whole manual serialization circus going on.

But once the PR is merged I think it would be an interesting thing to investigate separately to see if we can bring that property back. IIRC, when we were last looking at optimizing performance and looking at flamegraphs we _did_ spend a lot of relative effort cloning configurations.

.github/test-configuration.json
```diff
@@ -15,7 +15,7 @@
   },
   "ndc-postgres-12": {
     "feature": "postgres",
-    "filter": "not test(configure_initial_configuration_is_unchanged) and not test(configure_is_idempotent)",
+    "filter": "not test(postgres_current_only)",
```
plcplc, line 18: @danieljharvey : How does this look to you?

crates/connectors/ndc-postgres/src/configuration.rs
plcplc: Done!

crates/tests/databases-tests/src/postgres/configuration_tests.rs
```diff
@@ -1,32 +1,46 @@
 //! Tests that configuration generation has not changed.
 //!
 //! If you have changed it intentionally, run `just generate-chinook-configuration`.
+//!
```
danieljharvey, line 4: 👍 Yeah this should stop confusion in future, nice!

.github/test-configuration.json
danieljharvey: Looks great - naming it like this explains why we're skipping too which is nice.

PR #209, opened 2023-12-12
read-only transactions for query and support transaction isolation level configuration

soupi:
### What

- Allow users to configure the transaction isolation level via the configuration file.
- Run queries in read-only mode.

### How

1. change [simple_query_execution_plan](https://github.com/hasura/ndc-postgres/blob/main/crates/query-engine/sql/src/sql/execution_plan.rs#L52) to be wrapped in a transaction like in [simple_mutation_execution_plan](https://github.com/hasura/ndc-postgres/blob/main/crates/query-engine/sql/src/sql/execution_plan.rs#L92) (and add the relevant enums alternatives in ast.rs)
2. add [pre](https://github.com/hasura/ndc-postgres/blob/main/crates/query-engine/execution/src/mutation.rs#L62-L65) and [post](https://github.com/hasura/ndc-postgres/blob/main/crates/query-engine/execution/src/mutation.rs#L100-L102) statements to execution/query.rs as well
3. Add the isolation level configuration to v2 configuration and thread it to the parts that create the execution plan


crates/connectors/ndc-postgres/src/explain.rs
```diff
@@ -89,23 +89,27 @@ fn plan_query(
...
-                    state.metrics.error_metrics.record_unsupported_feature();
-                    connector::ExplainError::UnsupportedOperation(err.to_string())
-                }
-                _ => {
-                    state.metrics.error_metrics.record_invalid_request();
-                    connector::ExplainError::InvalidRequest(err.to_string())
-                }
+    let result = translation::query::translate(
+        &configuration.metadata,
+        &sql::ast::transaction::IsolationLevel::default(),
```
plcplc, line 94: A somewhat academic question perhaps: Can the transaction isolation level affect the query plan?

As a hot take I would think it wouldn't, but as a matter of being conservative it might be good to use the isolation level specified in the configuration still.

soupi: Fair point. I will add a new ticket for this.

soupi: Oh I will change this now, but we should probably run `EXPLAIN` in a transaction as well to make sure the postgres query plan is the same.

PR #210, opened 2023-12-12
chore: generate CRUD delete procedures

danieljharvey:
### What

We'd like to offer our users obvious mutations for free. We start by offering `delete_<tablename>_by_<some_unique_column>`.

`some_unique_column` must be a non-compound key (ie, `int` is fine). If they key is complex, we don't generate a procedure for it.

### How

We don't store the generated procedures in metadata currently, they are generated on the fly using `metadata.tables`. I don't think this is ideal but would rather make sure we're happy with the broad idea here before getting into that.

We generate them a) when the `/schema` endpoint is called, so we can tell the user what is available and b) when the user tries to run a procedure.


crates/query-engine/translation/src/translation/mutation/mod.rs
```diff
@@ -1,122 +1,3 @@
-//! Translate an incoming `MutationRequest`.
```
danieljharvey, line 1: Moved into `translate.rs` unchanged.

crates/query-engine/translation/src/translation/mutation/delete.rs
```diff
@@ -0,0 +1,90 @@
+use query_engine_metadata::metadata;
+use query_engine_sql::sql::ast;
+
+// this can get us `DELETE FROM <table> WHERE column = <column_name_arg>`
+// TODO: think about how RETURNING should work - ideally we'd not bake the columns in,
```
soupi, line 5: imo just put `RETURNING *`. The query part will deal with selecting the relevant columns.

crates/query-engine/translation/tests/common/mod.rs
```diff
@@ -62,7 +62,7 @@ pub fn test_mutation_translation(testname: &str) -> Result<String, translation::
         .operations
         .into_iter()
         .map(|operation| {
-            translation::mutation::translate(
+            translation::mutation::translate::translate(
```
soupi, line 65: nit: we could probably re-export this from `mod.rs`.

crates/connectors/ndc-postgres/src/schema.rs
```diff
@@ -233,6 +235,14 @@ pub async fn get_schema(
         })
         .collect();
 
+    let generated_procedures: Vec<models::ProcedureInfo> =
+        query_engine_translation::translation::mutation::generate::generate(&metadata.tables)
+            .iter()
+            .map(|(name, mutation)| mutation_to_procedure(name, mutation))
+            .collect();
+
+    procedures.extend(generated_procedures);
```
soupi, line 244: We should probably consider what to do on name clashes with user defined native query procedures.

crates/query-engine/translation/src/translation/mutation/delete.rs
```diff
@@ -0,0 +1,90 @@
...
+
+    #[test]
+    fn delete_to_sql() {
+        let delete = sample_delete();
+
+        let result = translate_delete(delete);
+
+        let mut sql = string::SQL::new();
+        result.to_sql(&mut sql);
+        insta::assert_json_snapshot!(sql.sql);
```
soupi, line 88: Would be really nice if possible to not have snapshots inside the `src/` dir (for example it could make `grep` and `sed` harder to use).

Also if you do:

```rs

    let pretty = sqlformat::format(
        &sql.sql,
        &sqlformat::QueryParams::None,
        sqlformat::FormatOptions::default(),
    );
    insta::assert_snapshot!(pretty);
```

The snapshot output will be nicer :)

crates/connectors/ndc-postgres/src/schema.rs
danieljharvey: Currently, any user-defined thing overrides the generated one. This seems the right way round. We can re-address this if we choose to store the generated stuff in metadata, in which case the user might manually rename things and we should help them not make confusing mistakes.

crates/query-engine/translation/src/translation/mutation/delete.rs
danieljharvey: Ahh yes, that is nicer. Personally, I like "tests being near the code" more than I care about where the snapshots end up, but can move the test if it's a real deal breaker.

crates/connectors/ndc-postgres/src/schema.rs
```diff
@@ -261,3 +271,43 @@ fn type_to_type(typ: &metadata::Type) -> models::Type {
...
+    match delete {
+        delete::DeleteMutation::DeleteByKey {
+            by_column,
+            description,
+            table_name: ast::TableName(table_name),
+            schema_name: ast::SchemaName(schema_name),
+        } => {
+            let mut arguments = BTreeMap::new();
+
+            let result_type = if schema_name == "public" {
```
danieljharvey, line 291: I dislike this, there must be a nicer way to get a matching type name for a collection.

crates/query-engine/translation/src/translation/helpers.rs
```diff
@@ -137,6 +137,21 @@ impl<'a> Env<'a> {
             .ok_or(Error::ProcedureNotFound(procedure_name.to_string()))
     }
 
+    pub fn lookup_generated_mutation(
+        &self,
+        procedure_name: &str,
+    ) -> Result<crate::translation::mutation::generate::Mutation, Error> {
+        // this means we generate them on every mutation request
```
danieljharvey, line 144: When I say "we generate them" I don't mean all the SQL for each, but we work out which mutations we CAN have based on the metadata.

crates/tests/databases-tests/src/citus/snapshots/databases_tests__citus__schema_tests__schema_test__get_schema.snap
```diff
@@ -3048,5 +3048,169 @@ expression: result
...
+            "name": "int4"
+          }
+        }
+      },
+      "result_type": {
+        "type": "named",
+        "name": "Employee"
+      }
+    },
+    {
```
danieljharvey, line 3119: This is how the generated mutations look in the schema output.

crates/tests/databases-tests/src/postgres/snapshots/databases_tests__postgres__mutation_tests__basic__delete_invoice_line.snap
```diff
@@ -0,0 +1,17 @@
+---
+source: crates/tests/databases-tests/src/postgres/mutation_tests.rs
+expression: result
+---
+{
```
danieljharvey, line 5: Look! It returns things!

crates/connectors/ndc-postgres/src/schema.rs
```diff
@@ -261,3 +271,43 @@ fn type_to_type(typ: &metadata::Type) -> models::Type {
         },
     }
 }
+
+fn mutation_to_procedure(name: &String, mutation: &generate::Mutation) -> models::ProcedureInfo {
```
danieljharvey, line 275: Given the description of a generated mutation, create a Procedure for it.

crates/query-engine/translation/src/translation/mutation/delete.rs
```diff
@@ -0,0 +1,111 @@
...
+pub enum DeleteMutation {
+    DeleteByKey {
+        description: String,
+        schema_name: ast::SchemaName,
+        table_name: ast::TableName,
+        by_column: metadata::database::ColumnInfo,
+    },
+}
+
+pub fn translate_delete(
```
danieljharvey, line 17: Given the description of a delete mutation (ie, `DeleteMutation`), and the arguments, output the SQL AST.

crates/query-engine/translation/src/translation/mutation/generate.rs
```diff
@@ -0,0 +1,75 @@
...
+use query_engine_sql::sql::ast;
+use std::collections::BTreeMap;
+
+#[derive(Debug, Clone)]
+pub enum Mutation {
+    DeleteMutation(DeleteMutation),
+}
+
+/// Given our introspection data, work out all the mutations we can generate
+pub fn generate(tables_info: &database::TablesInfo) -> BTreeMap<String, Mutation> {
```
danieljharvey, line 14: This is where all CRUD mutations will be generated.

crates/connectors/ndc-postgres/src/schema.rs
plcplc: Oh yeah. Which schemas are qualified or not depends on ndc raw config too.

The logic that decides which names are used for collection types should really be centralized so that we use the same everywhere we need that.

crates/query-engine/translation/src/translation/helpers.rs
plcplc: This is perfectly fine for a first version.

crates/query-engine/translation/src/translation/mutation/translate.rs
```diff
@@ -0,0 +1,253 @@
...
+                    env.lookup_generated_mutation(&name).and_then(|mutation| {
+                        translate_generated_mutation(&env, name, fields, arguments, mutation)
+                    })
+                }
+            }
+        }
+    }
+}
+
+fn translate_generated_mutation(
```
plcplc, line 46: The name of this function seems to be a bit off for what it does.

crates/query-engine/translation/src/translation/mutation/translate.rs
```diff
@@ -0,0 +1,253 @@
...
+        aggregates: Some(indexmap!("affected_rows".to_string() => models::Aggregate::StarCount{})),
+        fields: fields.clone(),
+        limit: None,
+        offset: None,
+        order_by: None,
+        predicate: None,
+    };
+
+    // return type of our mutation, there must be a nicer way to get the relevant collection type
+    // name
```
plcplc, line 77: Oh dear, one more?

plcplc (changes requested):
I think this is very nice.

But I think that, before we merge this there are two points that should be brought to order:

1. We should have a consistent way to derive collection type names from tables (as you too remark on)
2. We should (as a minimum) state explictly how this interacts with versioning and breaking changes. Especially since this PR is the first take on a feature undergoing evolution. (This could be as simple as stating that we'll bump the version number next time we change them, or we could add a configuration option for an opt-in feature preview, if we think we can manage that better than the "experimental-features" of v2)

crates/query-engine/translation/src/translation/mutation/translate.rs
danieljharvey: Agree, I don't think "generated" is a helpful word to bring with us into the codebase since nearly the whole thing generates something or other.

plcplc: Yeah - also it only produces delete mutations, right?

crates/query-engine/translation/src/translation/mutation/delete.rs
soupi: perhaps we can use [`insta::with_settings!`](https://docs.rs/insta/latest/insta/macro.with_settings.html) to [`set_snapshot_path`](https://docs.rs/insta/latest/insta/struct.Settings.html#method.set_snapshot_path) to the `tests/` dir? That way we can keep the tests near the code but the snapshots in a different place.

crates/connectors/ndc-postgres/src/schema.rs
danieljharvey: Ahh, so it turns out the introspection itself is the source of this logic, and uses it to generate the keys in the `metadata.tables` map, so using those instead of generating things.

crates/query-engine/sql/src/sql/rewrites/constant_folding.rs
```diff
@@ -94,6 +94,7 @@ pub fn normalize_cte(mut cte: CommonTableExpression) -> CommonTableExpression {
                 })
                 .collect(),
         ),
+        CTExpr::Delete(delete) => CTExpr::Delete(delete), // TODO: normalize this properly
```
soupi, line 97: This should be easy - normalize the where and mutate the delete. right?

danieljharvey: Oh ofc! Forgot this. Indeed, nothing too complicated I'd imagine.

danieljharvey: Done!

crates/tests/databases-tests/src/citus/snapshots/databases_tests__citus__schema_tests__schema_test__get_schema.snap
plcplc: :heart: @ `v1_delete_…`

crates/connectors/ndc-postgres/src/configuration/version2.rs
```diff
@@ -76,15 +76,12 @@ impl Default for ConfigureOptions {
             excluded_schemas: version1::default_excluded_schemas(),
             unqualified_schemas: version1::default_unqualified_schemas(),
             comparison_operator_mapping: version1::default_comparison_operator_mapping(),
-            mutations_version: None,
+            mutations_version: None, // we'll change this to `Some(MutationsVersions::V1)` when we
```
plcplc, line 79: :+1:

plcplc (approved):
I'm very fine with this.

soupi (commented):
Looks good! Just some docstring suggestions.

PR #214, opened 2023-12-18
support queries with no fields specified

soupi:
### What

We want to support queries with no fields so we can run queries such as:

```graphql
query {
  Albums(where: ..., limit: 10) {
    __typename
  }
}
```

### How

- We generate a query for rows without bailing even if the query does not contain fields (which generate a `SELECT FROM ...` query), but we will also indicate whether we got fields or not (None or an empty indexmap means no, otherwise yes).
- If the query specifies aggregates and we did not get row fields, we return the aggregate results fields only.
- If there are no aggregate fields specified, the query will always return rows result, which might be empty fields.
- We remove the NoFields error option.


crates/query-engine/translation/src/translation/query/mod.rs
```diff
@@ -85,32 +85,20 @@ pub fn translate_query(
...
+        root::translate_aggregate_query(env, state, current_table, from_clause, &query)?;
 
     match (row_select, aggregate_select) {
-        (Some(rows), None) => Ok(sql::helpers::SelectSet::Rows(rows)),
-        (None, Some(aggregates)) => Ok(sql::helpers::SelectSet::Aggregates(aggregates)),
-        (Some(rows), Some(aggregates)) => {
+        ((_, rows), None) => Ok(sql::helpers::SelectSet::Rows(rows)),
+        ((root::ReturnsFields::NoFieldsWereRequested, _), Some(aggregates)) => {
+            Ok(sql::helpers::SelectSet::Aggregates(aggregates))
+        }
```
plcplc, line 99: Do we know that `fields` and `aggregates` are mutually exclusive?

Is it conceivable that a user could have issued a query of `__typename` alongside a set of aggregates?

plcplc (approved):
Good stuff. But it'd be good to clear up if there is a potential issue wrt fields vs aggregates (see the comment below).

crates/query-engine/translation/src/translation/query/mod.rs
plcplc: That is a fact, certainly, though it strikes me as a bit weird. (I don't mind leaving it be though)

PR #216, opened 2023-12-18
prepare for v0.2

soupi:
### What

We've made a few cool updates that would be nice to release as v0.2.0. This PR aligns the changelog for release.


changelog.md
```diff
@@ -2,17 +2,20 @@
...
-- Support array types ([#191](https://github.com/hasura/ndc-postgres/pull/191), ...)
+- Support array types ([#191](https://github.com/hasura/ndc-postgres/pull/191))
 - Support Native Query Mutations ([#189](https://github.com/hasura/ndc-postgres/pull/189), [#198](https://github.com/hasura/ndc-postgres/pull/198))
 
 ## [0.1.0] - 2023-11-29
 
 Initial release.
 
-[Unreleased]: https://github.com/hasura/ndc-postgres/compare/v0.1.0...HEAD
+[Unreleased]: https://github.com/hasura/ndc-postgres/compare/v0.2.0...HEAD
```
plcplc, line 19: Hm.. why is this version list at the bottom of the file? It looks like this PR is making an addition to v0.1.0.

Wouldn't it be better to have the list at the top of the file? Sort of like a table of contents?

plcplc (approved):

changelog.md
soupi: This is not a list. This is markdown syntax for links that appear somewhere in the document (so you can write `[Unreleased]` and it refers to a link to this link.
Render this changelog and see :) https://github.com/hasura/ndc-postgres/blob/24010c16f5a9b41655989c9d2dde097b9fad91b7/changelog.md

plcplc: Ah, it seems like a simplified version of the footnote syntax. 

How neat - I didn't know about that.

Let's keep that then!

changelog.md
```diff
@@ -2,10 +2,13 @@
 
 ## [Unreleased]
 
+## [0.2.0] - 2023-12-20
```
soupi, line 5:
```suggestion
## [0.2.0] - 2023-12-21
```

PR #219, opened 2023-12-21
fix `new-configuration.sh` on mac

soorajshankar:
### What

The users who try to run V3 OSS engine with NDC-Postgres were following this [readme](https://github.com/hasura/graphql-engine/tree/master/v3#readme), and it was using the new-configuration.sh file to generate the pg schema. 

However this script was not working on mac shell runtime as the `EXTRA_CONFIG="${3:-{\}}"` was defaulting to {/} in mac instead of {}. 

This PR fixes that and default to the {}

### How

Added default value as an extra variable and its now working on both linux and mac.


soupi (approved):
Thanks!

PR #220, opened 2023-12-21
Prefix function model and translation support

plcplc:
### What

This PR adds the ability use prefix-functions (as opposed to infix operators) as comparison operators.

It does not add any introspection logic.

### How

The comparison operator metadata configuration now includes a boolean field `isInfix, distinguishing prefix functions from infix operator.

By volume, most of the diff of this PR deals with adding `"isInfix": true` assignments to test deployment data.


crates/query-engine/translation/tests/snapshots/tests__it_select_where_prefix_function.snap
```diff
@@ -0,0 +1,27 @@
...
+        SELECT
+          coalesce(json_agg(row_to_json("%2_rows")), '[]') AS "rows"
+        FROM
+          (
+            SELECT
+              "%0_Album"."AlbumId" AS "Id"
+            FROM
+              "public"."Album" AS "%0_Album"
+            WHERE
+              some_prefix_function("%0_Album"."AlbumId", 10)
```
soupi, line 22: in general, a real function is preferable because you can copy paste the sql into repl-postgres and get a result

soupi (approved):
:ok_hand:

PR #223, opened 2023-12-21
Prefix function introspection

plcplc:
### What

This PR adds support for introspecting prefix-functions (i.e. functions which are not infix operators) as comparison operators.

### How

There are a myriad of functions defined in Postgres that have a type compatible with that of a comparison operator, but which are not useful as such, or which are simply prefix-function versions of the more well know infix operators.

For instance, `has_schema_privilege' (et. al.) interrogates the server's user authorization rules, and `texteq` is simply `(=)` specialized to the `text` type.

Since we don't want to clutter up the default introspected metadata with at best useless and at worst insecure functions we provide the new configuration option `introspectPrefixFunctionComparisonOperators` which lists out the functions that will be selected for introspection. By default this list includes various functions on builtin Postgres types and the types of PostGIS.


soupi (approved):

soupi:
subsumed by #220 #221 and #223.

PR #237, opened 2024-01-02
test that failed mutations are rolled back

soupi:
### What

We want to test that mutation requests are run as a transaction, and that if one mutation fails, the other mutations are rolled back.

### How

We add a test with a mutation request that tries to insert an album with an already existing id, which fails. We check that it fails, and we also verify that the artist we inserted before was not inserted to the db.

### DIFF
--- a/crates/tests/databases-tests/src/postgres/snapshots/databases_tests__postgres__mutation_tests__basic__insert_artist_album.snap
+++ b/crates/tests/databases-tests/src/postgres/snapshots/databases_tests__postgres__mutation_tests__basic__insert_artist_album.snap
@@ -2,40 +2,51 @@
 source: crates/tests/databases-tests/src/postgres/mutation_tests.rs
 expression: result
 ---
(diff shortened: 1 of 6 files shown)


crates/tests/databases-tests/src/postgres/snapshots/databases_tests__postgres__mutation_tests__basic__insert_artist_album.snap
```diff
@@ -2,40 +2,51 @@
...
-          "__value": [
-            {
-              "artist_id": 276,
-              "name": "Olympians"
-            }
-          ]
-        }
-      ]
-    },
+[
```
soupi, line 5: we modified the existing test to also run a select after the mutation to make sure that the insert worked. We put the two (separate) requests in a tuple, which is why this is now an array.

i-am-tom (approved):
LGTM

PR #239, opened 2024-01-04
Better error codes

soupi:
### What

Currently we return 500 internal error whenever there's an error.
Some errors should return different errors such as 400 or 422.

### How

See the changes to the code in `crates/connectors/src/*.rs` to see which error is mapped to which ndc-hub error.

We also catch specific errors from postgres now, if the code in of the `22` class we return a 422 error, if it is from the `23` class we return a 403 error (for mutations).


plcplc (approved):
Looks good to me.

PR #240, opened 2024-01-05
Support composite types: models and translation

plcplc:
### What

This PR adds limited support for composite types.

The ndc spec does not provide any means to interact with a composite/object type other than allowing them to appear in results and as input arguments. It does not e.g. let you perform field selection or comparison operators. This PR supports that.

What is not yet supported is the introspection of composite types from the database. This will appear in a follow-up PR.


crates/query-engine/translation/tests/goldenfiles/select_composite_column/tables.json
soupi, line 1: It would be nice if we could have a meaningful use case here, `things`, `thing_a`, `thing_b`, `thing1` is all very confusing together.

crates/query-engine/translation/src/translation/query/values.rs
```diff
@@ -38,6 +38,25 @@ pub fn translate_json_value(
                 r#type: type_to_ast_scalar_type(r#type),
             })
         }
+        // For composite types we use the `jsonb_populate_record` function to build the composite
+        // value rather than constructing the value field-by-field in SQL.
+        (serde_json::Value::Object(_obj), database::Type::CompositeType(_type_name)) => {
+            let stringified = serde_json::to_string(value)
```
soupi, line 44: Worth considering passing a serde_json::Value instead of a string here.

soupi (approved):

crates/query-engine/translation/src/translation/query/values.rs
plcplc: Oh! I didn't know we could!

PR #241, opened 2024-01-05
Return the generated SQL of an explain request with empty variables

soupi:
### What

To support use cases such as explaining remote-relationships, we want to be able to view the generated SQL of a query that uses variables but does not provide variables (since those would need to be run first in order to be provided).
In this PR we allow explain requests on queries that provide empty variables (`variables: []`) to receive just the generated SQL, without the postgres explain part.

### How

We check that we receive an empty variables field and if that's the case we will skip querying postgres and just return the generated SQL.

Please read this PR with the `hide whitespace` setting.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -6,6 +6,13 @@
 
 ### Added
 
+- Return the generated SQL on an explain request with empty variables
+  ([#241](https://github.com/hasura/ndc-postgres/pull/241))
+- Emit invalid request, constraint not met, and unprocessable content errors at the relevant scenarios
+  ([#239](https://github.com/hasura/ndc-postgres/pull/239))
+
+### Added
+
 - Support for introspecting prefix-functions as comparison operators ([#223](https://github.com/hasura/ndc-postgres/pull/223))
 - Support prefix functions as comparison operators ([#220](https://github.com/hasura/ndc-postgres/pull/223))
 - Support queries without fields specified. ([#209](https://github.com/hasura/ndc-postgres/pull/209))
--- a/changelog.md
+++ b/changelog.md
@@ -6,7 +6,7 @@
 
 ### Added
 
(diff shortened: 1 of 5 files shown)


changelog.md
```diff
@@ -6,6 +6,13 @@
 
 ### Added
 
+- Return the generated SQL on an explain request with empty variables
+  ([#241](https://github.com/hasura/ndc-postgres/pull/241))
+- Emit invalid request, constraint not met, and unprocessable content errors at the relevant scenarios
```
soupi, line 11: forgot to add this for my previous pr.

changelog.md
```diff
@@ -6,6 +6,13 @@
 
 ### Added
 
+- Return the generated SQL on an explain request with empty variables
```
soupi, line 9:
```suggestion
- Return the generated SQL of an explain request with empty variables
```

soupi:
> But shouldn't we be able to still explain the query if we just pass in in the value [] for the variables parameter?

PG will throw an error in that case:

```
error returned from database: column %0_%variables_table.search does not exist
```

plcplc:
Hm.., are we sure about that? This small test doesn't seem to have any issue with a `[]`:

```
postgres=# select * from json_to_recordset('[]') as t(id int, foo text);
 id | foo
----+-----
(0 rows)

```

Am I maybe missing some detail?

plcplc:
Ah, I think I got it. We base our idea of the shape of the variables table on the variables that were in fact supplied rather than which ones the query refers to. And so we don't know what variables ought to exist when none are given.

plcplc:
I agree we should just do it the way of this PR as a matter of expediency. But it would still be nice to have the query inform the variables rather than the other way around.

soupi:
We can think about it when it becomes a requirement :)

PR #249, opened 2024-01-15
Support variables of composite type

plcplc:
### What

This PR changes how the `variables` field of a request is translated, such that variables of composite types are now supported.

Array types require a some treatment of their own, which will be dealt with in a follow-up PR.

### How

Prior to this PR, incoming variables (bound to `$1` in this example) were translated in the style of:

```
  ... FROM json_to_recordset($1) as "%variables_table"("%variable_order" int, "variable1" varchar, "variable2" varchar, "variable3" varchar, ...)
```

All the variables used in the request (`variable1`, `variable2`, etc) all become columns of `%variables_table`, and all of type `varchar`.

Usage sites would rely on casts to interpret the variables as the relevant scalar type:

```
 cast("%variables_table"."variable1" as int4)
```

On this approach we can observe a few things:

* We can only support plain scalar types, since we cannot canonically represent an array or object as a string through `json_to_recordset`.
* "%variable_order" is a reserved name that can conflict with user variable names.
* We don't need to do any bookkeeping of what type a variable has. Usage context takes care of this to the extent that queries without type errors execute correctly.

In order to support composite (and array) types, we change the translation scheme:

```
  ... FROM json_to_recordset($1) as "%variables_table"("%variable_order" int, "%variables" jsonb)
```

All the variables used in the request (`variable1`, `variable2`, etc) all become fields of the "%variables" column, which remains yet a json value.

Usage sites still rely on casts to interpret the variables as the relevant scalar type, but now have to rely on the `->>` operator to project a variable from the `%variables` object as a plain `text` scalar:

```
 cast("%variables_table"."%variables" ->> "variable1" as int4)
```

In the case of composite types we use the function `jsonb_populate_record` to construct the composite value from the json-valued variable-field, and the operator `->` to extract the field value as `jsonb` (c.f. `->>` for `text`)

```
jsonb_populate_record(null::person, "%variables_table"."%variables" -> "variable2")
```

On this updated approach we can observe a few things:

* We can support both scalars, arrays, and composite types.
* "%variable_order" is no longer a reserved name.
* We still don't need to do any bookkeeping of what type a variable has.
* `jsonb_populate_record` does all the heavy lifting, which means that we don't need to generate more than a single call to this function to handle any level of nested types.

### DIFF
--- a/crates/connectors/ndc-postgres/src/query.rs
+++ b/crates/connectors/ndc-postgres/src/query.rs
@@ -93,10 +93,6 @@ async fn execute_query(
                 tracing::error!("{}", err);
                 // log error metric
                 match &err {
-                    query_engine_execution::query::QueryError::ReservedVariableName(_) => {
-                        state.metrics.error_metrics.record_invalid_request();
-                        connector::QueryError::InvalidRequest(err.to_string())
-                    }
                     query_engine_execution::query::QueryError::VariableNotFound(_) => {
                         state.metrics.error_metrics.record_invalid_request();
                         connector::QueryError::InvalidRequest(err.to_string())
--- a/crates/query-engine/execution/src/query.rs
+++ b/crates/query-engine/execution/src/query.rs
@@ -248,8 +248,8 @@ fn variables_to_json(
 
                 let variables_field = serde_json::Value::Object(
                     varset
-                        .clone()
-                        .into_iter()
+                        .iter()
+                        .cloned()
                         .collect::<serde_json::Map<String, serde_json::Value>>(),
                 );
                 row.insert(sql::helpers::VARIABLES_FIELD.to_string(), variables_field);
--- a/crates/query-engine/execution/src/query.rs
+++ b/crates/query-engine/execution/src/query.rs
@@ -248,8 +248,8 @@ fn variables_to_json(
 
                 let variables_field = serde_json::Value::Object(
                     varset
-                        .iter()
-                        .cloned()
+                        .clone()
+                        .into_iter()
                         .collect::<serde_json::Map<String, serde_json::Value>>(),
                 );
                 row.insert(sql::helpers::VARIABLES_FIELD.to_string(), variables_field);
--- a/crates/query-engine/execution/src/query.rs
+++ b/crates/query-engine/execution/src/query.rs
@@ -239,20 +239,22 @@ fn variables_to_json(
             .iter()
             .enumerate()
             .map(|(i, varset)| {
-                let mut varset = varset
-                    .clone()
-                    .into_iter()
-                    .collect::<serde_json::Map<String, serde_json::Value>>();
-                match varset.insert(
+                let mut row = serde_json::Map::new();
+
+                row.insert(
                     sql::helpers::VARIABLE_ORDER_FIELD.to_string(),
                     serde_json::Value::Number(i.into()),
-                ) {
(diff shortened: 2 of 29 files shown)


crates/connectors/ndc-postgres/src/query.rs
```diff
@@ -93,10 +93,6 @@ async fn execute_query(
                 tracing::error!("{}", err);
                 // log error metric
                 match &err {
-                    query_engine_execution::query::QueryError::ReservedVariableName(_) => {
```
plcplc, line 96: There are no longer any occasions that can result in a variable name being reserved.

crates/query-engine/execution/src/query.rs
```diff
@@ -239,20 +239,22 @@ fn variables_to_json(
...
-                Ok(serde_json::Value::Object(varset))
+                );
+
+                let variables_field = serde_json::Value::Object(
+                    varset
+                        .clone()
+                        .into_iter()
+                        .collect::<serde_json::Map<String, serde_json::Value>>(),
+                );
+                row.insert(sql::helpers::VARIABLES_FIELD.to_string(), variables_field);
```
plcplc, line 255: Construct the `(%variable_order int, %variables jsonb)` table.

crates/query-engine/sql/src/sql/convert.rs
```diff
@@ -166,12 +166,12 @@ impl From {
                 sql.append_syntax(" AS ");
                 alias.to_sql(sql);
             }
-            From::JsonToRecordset {
+            From::JsonbToRecordset {
```
plcplc, line 169: We use `jsonb` instead of `json` because json-valued query parameters appear to be bound as `jsonb`, which means we avoid having to cast (see the updated test snapshots for examples)

crates/tests/databases-tests/src/cockroach/query_tests.rs
```diff
@@ -45,6 +45,12 @@ mod basic {
         let result = run_query(create_router().await, "native_queries/embedded_variable").await;
         insta::assert_json_snapshot!(result);
     }
+
+    #[tokio::test]
+    async fn select_composite_variable() {
+        let result = run_query(create_router().await, "select_composite_variable").await;
```
plcplc, line 51: TODO: Leave out this test, since cockroach does not support nested user defined types.

soupi (approved):
Looks great. Really appreciate the thorough PR description, it really helped me review this PR quickly.

crates/query-engine/sql/src/sql/helpers.rs
```diff
@@ -558,40 +557,23 @@ pub fn select_row_as_json_with_default(
 ///
 /// ```sql
 /// FROM
-///   json_to_recordset(cast('[{"%variable_order": 1, "search": "%Good%"}]' as json))
-///     AS "%0_variables"("search" varchar, "%variable_order" int)
+///   json_to_recordset(cast('[{"%variable_order": 1, "%variables": {"search": "%Good%", ...}}]' as json))
+///     AS "%0_variables"("%variables" json, "%variable_order" int)
```
SamirTalwar, line 561:
```suggestion
///     AS "%0_variables"("%variable_order" int, "%variables" jsonb)
```

SamirTalwar:
You need to run `cargo build` to generate the new lockfile.

PR #254, opened 2024-01-15
Replace use of "deployment" with "ndc metadata"

i-am-tom:
I (specifically Tom) keep getting confused about whether deployments are for engine or the NDC bit, and the first time I drafted this PR @plcplc actually had to point out that I got the difference wrong _again_, so this PR replaces mentions of "deployment" with "ndc metadata" to make it more obvious to me (specifically Tom).


scripts/archive-old-deployment.sh
i-am-tom, line 1: This is now called scripts/archive-old-ndc-metadata.sh

crates/tests/tests-common/src/deployment/configuration.rs
i-am-tom, line 1: This is now called crates/tests/tests-common/src/ndc_metadata/configuration.rs

crates/tests/tests-common/src/deployment/mod.rs
i-am-tom, line 1: This is now called crates/tests/tests-common/src/ndc_metadata/mod.rs

crates/tests/databases-tests/src/deployment_snapshot_tests.rs
i-am-tom, line 1: This is now called crates/tests/databases-tests/src/ndc_metadata_snapshot_tests.rs

SamirTalwar (approved):

PR #256, opened 2024-01-16
introspect columns to include default, generated and identity columns

soupi:
### What

In order to generate insert mutations we'd like to know which columns should be inserted into, which we shouldn't insert to, and which can go either way.
To do that we add settings for default, generated, and identity information to column information, as well as introspection code.

### How

1. We create an enum for each setting, and include that information in `ColumnInfo` metadata
2. We add SQL introspection and query this information from `pg_attribute`, but see the comment about generated columns


crates/query-engine/metadata/src/metadata/database.rs
```diff
@@ -79,10 +107,31 @@ pub struct ColumnInfo {
...
+    pub is_identity: IsIdentity,
+    #[serde(skip_serializing_if = "is_not_generated")]
+    #[serde(default)]
+    pub is_generated: IsGenerated,
     #[serde(default)]
     pub description: Option<String>,
 }
 
+fn does_not_have_default(has_default: &HasDefault) -> bool {
+    matches!(has_default, HasDefault::NoDefault)
```
plcplc, line 124: TIL

soupi: It was a clippy suggestion :)

crates/connectors/ndc-postgres/src/configuration/version2.sql
```diff
@@ -102,11 +102,17 @@ WITH
...
+      CASE WHEN att.atthasdef THEN 'hasDefault' ELSE 'noDefault' END
+      AS has_default,
+      CASE WHEN att.attidentity = 'd' THEN 'identityByDefault'
+           WHEN att.attidentity = 'a' THEN 'identityAlways'
+           ELSE 'notIdentity'
+      END
+      AS is_identity
+      -- skipped because yugabyte is based on pg11 that does not have this field.
+      -- CASE WHEN att.attgenerated = 's' THEN 'isGenerated' ELSE 'notGenerated' END
+      -- AS is_generated
```
plcplc, line 115: There is a trick being employed in `pg_table_metadata.sql` of HGE v2 which deals with this very issue. It could well be worth it to port that over.

plcplc (approved):

crates/connectors/ndc-postgres/src/configuration/version2.sql
soupi: Ah good to know, for reference: https://github.com/hasura/graphql-engine-mono/blob/main/server/src-rsr/pg_table_metadata.sql#L70-75

Will do that in a follow-up PR.

PR #257, opened 2024-01-16
Add more tests of composite types 

plcplc:
### What

* Cockroach does support non-nested user defined types, so we test those as well
* Yugabyte is now also covered by tests of composite types
* Aurora is (tentatively) covered. The test database will be updated in a follow-up PR

### How

We now distinguish between "simple" and "complex" use of composite types, the difference being that the "complex" uses nest user defined record types.

The existing tests were of the "complex" type, and these are now named as such. This PR adds "simple" test cases that only deal with user defined record types that reference only scalar types.


soupi (approved):
nice

PR #258, opened 2024-01-16
Derive `Copy` for `IsolationLevel` so we don't have to clone it.

SamirTalwar:
### What

Just making the code more Rust-y.

It's a little thing but it got in my way earlier.

### How

By adding `Copy` to `IsolationLevel`, we can freely pass it around by value, not by reference.


soupi (approved):
Cool.

PR #261, opened 2024-01-17
Auto-generated insert procedures v1

soupi:
### What

We'd like to auto-generate insert procedures for tables.

### How

Soon, tl;dr:

1. We add the relevant INSERT INTO sql structure to the relevant parts
2. We generate insert procedures from the metadata information about tables and columns
3. We expose in the scheme by creating a procedure for each generated insert procedure, and create an object type for the table's insertable object
4. We add some custom-tables.sql to postgres so we can test other stuff not exposed by chinook

todo: add engine metadata for `v1_insert_custom_dog`.


crates/connectors/ndc-postgres/src/configuration/version2.sql
```diff
@@ -109,12 +109,14 @@ WITH
...
       END
-      AS is_identity
-      -- skipped because yugabyte is based on pg11 that does not have this field.
-      -- CASE WHEN att.attgenerated = 's' THEN 'isGenerated' ELSE 'notGenerated' END
-      -- AS is_generated
+      AS is_identity,
+      CASE WHEN attgenerated_exists
+	       THEN CASE WHEN attgenerated::text = 's' THEN 'isGenerated' ELSE 'notGenerated' END
+           ELSE 'notGenerated'
+      END as is_generated
```
SamirTalwar, line 116: Seems like a strange way to implement booleans. Why not `true` and `false`?

crates/connectors/ndc-postgres/src/schema.rs
```diff
@@ -252,16 +252,18 @@ pub async fn get_schema(
...
 
+    let mut more_object_types = BTreeMap::new();
     let generated_procedures: Vec<models::ProcedureInfo> =
         query_engine_translation::translation::mutation::generate::generate(
             &metadata.tables,
             &config.mutations_version,
         )
         .iter()
-        .map(|(name, mutation)| mutation_to_procedure(name, mutation))
+        .map(|(name, mutation)| mutation_to_procedure(name, mutation, &mut more_object_types))
```
SamirTalwar, line 262: I am really not a fan of a `map` which mutates. I suggest you either return procedures and types, aggregating them in `unzip`.

Nothing reads from the value, as far as I can tell, so this also affords way more control than is necessary.

crates/query-engine/translation/src/translation/mutation/insert.rs
```diff
@@ -0,0 +1,141 @@
...
+    pub collection_name: String,
+    pub description: String,
+    pub schema_name: ast::SchemaName,
+    pub table_name: ast::TableName,
+    pub columns: BTreeMap<String, metadata::database::ColumnInfo>,
+}
+
+/// generate an insert mutation.
+pub fn generate(
+    collection_name: &String,
```
SamirTalwar, line 24:
```suggestion
    collection_name: &str,
```

SamirTalwar (commented):
The implementation seems fine, I just have a couple of nitpicks.

crates/connectors/ndc-postgres/src/schema.rs
soupi: > I am really not a fan of a map which mutates.

Why not?

> I suggest you either return procedures and types, aggregating them in unzip.

I actually tried it first and it was very clumsy, which is why I moved to mutating.

crates/connectors/ndc-postgres/src/configuration/version2.sql
soupi: Just liked being explicit, wouldn't mind changing it to bool if you want me to.

crates/connectors/ndc-postgres/src/schema.rs
SamirTalwar: Yeah, I just tried it too and you're right, it's a bit too clunky.

Makes me sad, but 🤷🏾

soupi: Yeah I'd probably avoid mutation if this was nicer, but personally I feel like using mutation when it's localized is absolutely fine.

SamirTalwar: I will just note that this has been solved above by iterating twice.

Not sure if that's better or worse.

crates/connectors/ndc-postgres/src/configuration/version2.sql
soupi: I've changed this so instead of returning `isGenerated`, it will return `stored`, so that if and when postgres adds additional values to this, it will be strictly additive.

static/deployment-snapshots/d37aa40e768d774850e1fc638eab23559ef47a82f0f7f8acc44fe84fc4c9b0e2.json
soupi, line 1: a v1 config snapshot! Generated because we added a new table (custom.dog).

i-am-tom (approved):
Looks good to me

PR #262, opened 2024-01-17
Move variables_table to Env

plcplc:
### What

Since the table reference that binds the variables table cannot change in the course of query translation we move this to `Env` instead of `State`.


SamirTalwar (approved):
Looks sensible to me.

PR #263, opened 2024-01-19
Support variables of array type

plcplc:
### What

This PR extends the support for query variables to include arrays.

### How

We need to go a bit out of our way to be able to translate JSON arrays to Postgres arrays, since there is no single builtin function that does this.

As an example taken from the tests, a variable of type array-of-`organization` gets translated to the below SQL:

```
(
  SELECT
    array_agg(
      jsonb_populate_record(cast(null as organization), "%0_arr"."elem")
    ) AS "elem"
  FROM
    jsonb_array_elements(("%0_%variables_table"."%variables" -> $2)) AS "%0_arr"("elem")
)

```

Which, being a single-element set-returning expression, can feature both in a `FROM` clause and in a select-list or other expression.

### DIFF
--- a/crates/query-engine/translation/src/translation/query/native_queries.rs
+++ b/crates/query-engine/translation/src/translation/query/native_queries.rs
@@ -38,6 +38,15 @@ pub fn translate(env: &Env, state: State) -> Result<Vec<sql::ast::CommonTableExp
                             models::Argument::Variable { name } => match &variables_table {
                                 Err(err) => Err(err.clone()),
                                 Ok(variables_table) => Ok(values::translate_variable(
+                                    // We need a 'State' value when translating variables in order
+                                    // to be able to generate fresh names for bound relational
+                                    // expressions.
+                                    // However, we cannot readily re-use the 'state' argument we're
+                                    // given, since that is an "owned" value which is "moved" in
+                                    // the course of iterating over the native queries accumulated
+                                    // within.
+                                    // Ideally we should resolve this by tracking native queries
+                                    // separately from the fresh table name counter.
                                     &mut State::new(),
                                     variables_table.clone(),
                                     name.clone(),
--- a/crates/query-engine/translation/src/translation/query/native_queries.rs
+++ b/crates/query-engine/translation/src/translation/query/native_queries.rs
@@ -38,6 +38,7 @@ pub fn translate(env: &Env, state: State) -> Result<Vec<sql::ast::CommonTableExp
                             models::Argument::Variable { name } => match &variables_table {
                                 Err(err) => Err(err.clone()),
                                 Ok(variables_table) => Ok(values::translate_variable(
+                                    &mut State::new(),
                                     variables_table.clone(),
                                     name.clone(),
                                     &typ,
--- a/crates/query-engine/translation/src/translation/query/values.rs
+++ b/crates/query-engine/translation/src/translation/query/values.rs
@@ -94,28 +94,27 @@ pub fn translate_variable(
         column: sql::helpers::make_column_alias(sql::helpers::VARIABLES_FIELD.to_string()),
     });
 
-    let projected_variable_exp = match r#type {
-        database::Type::CompositeType(_) | database::Type::ArrayType(_) => {
-            sql::ast::Expression::BinaryOperation {
-                left: Box::new(variables_reference),
-                operator: sql::ast::BinaryOperator("->".to_string()),
-                right: Box::new(sql::ast::Expression::Value(sql::ast::Value::String(
-                    variable.clone(),
-                ))),
-            }
-        }
-        database::Type::ScalarType(_) => sql::ast::Expression::BinaryOperation {
-            left: Box::new(variables_reference),
-            operator: sql::ast::BinaryOperator("->>".to_string()),
-            right: Box::new(sql::ast::Expression::Value(sql::ast::Value::String(
-                variable.clone(),
-            ))),
-        },
+    // We use the binop '->' to project (as jsonb) the value of a variable from the data column of
+    // the variable table.
+    let projected_variable_exp = sql::ast::Expression::BinaryOperation {
+        left: Box::new(variables_reference),
+        operator: sql::ast::BinaryOperator("->".to_string()),
+        right: Box::new(sql::ast::Expression::Value(sql::ast::Value::String(
+            variable.clone(),
+        ))),
     };
 
     translate_projected_variable(state, r#type, projected_variable_exp)
 }
 
+/// Produce a SQL expression that translates an expression of Postgres type 'jsonb' into a given
+/// type.
+///
+/// For scalar types and object types this is a simple operation, since we can rely on builtin
+/// functions.
+///
+/// Arrays are more complex since there isn't a builtin function that handles array
+/// types.
 pub fn translate_projected_variable(
     state: &mut State,
     r#type: &database::Type,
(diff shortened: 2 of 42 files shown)


crates/query-engine/translation/src/translation/query/values.rs
```diff
@@ -84,48 +84,88 @@ fn type_to_ast_scalar_type(typ: &database::Type) -> sql::ast::ScalarType {
...
+    let variables_reference = Expression::ColumnReference(ColumnReference::AliasedColumn {
+        table: variables_table,
+        column: sql::helpers::make_column_alias(sql::helpers::VARIABLES_FIELD.to_string()),
+    });
 
-    match r#type {
-        database::Type::CompositeType(_type_name) => {
-            let exp = sql::ast::Expression::BinaryOperation {
+    let projected_variable_exp = match r#type {
+        database::Type::CompositeType(_) | database::Type::ArrayType(_) => {
```
soupi, line 98: Could you please add a comment of the intention of the difference between this block and the next block? Something along the lines of:

```suggestion
        // For composite types and array types, we use `->` to return the value as `json`
        // because ...
        database::Type::CompositeType(_) | database::Type::ArrayType(_) => {
```

crates/query-engine/translation/src/translation/query/values.rs
```diff
@@ -84,48 +84,88 @@ fn type_to_ast_scalar_type(typ: &database::Type) -> sql::ast::ScalarType {
...
-        _ => {
-            let exp = sql::ast::Expression::BinaryOperation {
-                left: Box::new(variables_reference),
-                operator: sql::ast::BinaryOperator("->>".to_string()),
-                right: Box::new(sql::ast::Expression::Value(sql::ast::Value::String(
-                    variable.clone(),
-                ))),
+        database::Type::ScalarType(_) => sql::ast::Expression::BinaryOperation {
+            left: Box::new(variables_reference),
+            operator: sql::ast::BinaryOperator("->>".to_string()),
```
soupi, line 109:
```suggestion
            // For scalar types, we use `->>` to return the value as `text`
            // because ...
            operator: sql::ast::BinaryOperator("->>".to_string()),
```

crates/query-engine/translation/src/translation/query/values.rs
```diff
@@ -84,48 +84,88 @@ fn type_to_ast_scalar_type(typ: &database::Type) -> sql::ast::ScalarType {
...
+            right: Box::new(sql::ast::Expression::Value(sql::ast::Value::String(
+                variable.clone(),
+            ))),
+        },
+    };
+
+    translate_projected_variable(state, r#type, projected_variable_exp)
+}
+
+pub fn translate_projected_variable(
```
soupi, line 119: Please add a docstring here :)

crates/query-engine/translation/src/translation/query/values.rs
```diff
@@ -84,48 +84,88 @@ fn type_to_ast_scalar_type(typ: &database::Type) -> sql::ast::ScalarType {
...
+            args: vec![
+                sql::ast::Expression::Cast {
+                    expression: Box::new(sql::ast::Expression::Value(sql::ast::Value::Null)),
+                    r#type: type_to_ast_scalar_type(r#type),
+                },
+                exp,
+            ],
+        },
+        database::Type::ArrayType(type_name) => {
+            let arr_table = state.make_table_alias("arr".to_string());
```
soupi, line 136: nit: can we please avoid abbreviations in code and names? lets use:
- `let array_table`
- `"array".to_string`
- `let element_column`
- `name: "element".to_string`
- `expression: `
- `element_expression`

soupi (commented):
Looks good overall. Had a few nits that are mostly about comments.

crates/query-engine/translation/src/translation/query/values.rs
plcplc: Sensible! I'll do that.

crates/query-engine/translation/src/translation/query/values.rs
```diff
@@ -84,48 +84,88 @@ fn type_to_ast_scalar_type(typ: &database::Type) -> sql::ast::ScalarType {
...
+                    function: sql::ast::Function::Unknown("array_agg".to_string()),
+                    args: vec![converted_element_exp],
+                },
+            )]);
+
+            result_select.from = Some(from_arr);
+
+            sql::ast::Expression::CorrelatedSubSelect(Box::new(result_select))
         }
+        _ => sql::ast::Expression::Cast {
```
SamirTalwar, line 166: If we can avoid catch-all patterns, let's do that.

```suggestion
        database::Type::ScalarType(_type_name) => sql::ast::Expression::Cast {
```

SamirTalwar (approved):
Pretty happy, just some minor comments.

crates/query-engine/translation/src/translation/query/values.rs
plcplc: These parts should be better commented now.

I also simplified the process by only using the `->` operator and moving the textual extraction to the `ScalarType`-case, which means that the function only assumes that th expression it's working on is a `jsonb`-value, and not contextually dependent on whether it's translating a scalar type or composite/array type.

crates/query-engine/translation/src/translation/query/native_queries.rs
```diff
@@ -38,6 +38,7 @@ pub fn translate(env: &Env, state: State) -> Result<Vec<sql::ast::CommonTableExp
                             models::Argument::Variable { name } => match &variables_table {
                                 Err(err) => Err(err.clone()),
                                 Ok(variables_table) => Ok(values::translate_variable(
+                                    &mut State::new(),
```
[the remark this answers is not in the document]
plcplc: done

crates/query-engine/translation/src/translation/query/values.rs
plcplc: done.

crates/query-engine/translation/src/translation/query/values.rs
```diff
@@ -84,48 +84,98 @@ fn type_to_ast_scalar_type(typ: &database::Type) -> sql::ast::ScalarType {
...
+            function: sql::ast::Function::JsonbPopulateRecord,
+            args: vec![
+                sql::ast::Expression::Cast {
+                    expression: Box::new(sql::ast::Expression::Value(sql::ast::Value::Null)),
+                    r#type: type_to_ast_scalar_type(r#type),
+                },
+                exp,
+            ],
+        },
+        database::Type::ArrayType(type_name) => {
```
soupi, line 134:
```suggestion
        // We translate projection of array types into the following sql:
        // ```
        // ( SELECT
        //     array_agg(
        //       jsonb_populate_record(cast(null as <type>), "array"."element")
        //     ) AS "element"
        //   FROM
        //    jsonb_array_elements((<variable_table> -> <label>)) AS "array"("element")
        // )
        // ```
        database::Type::ArrayType(type_name) => {
```

crates/query-engine/translation/src/translation/query/native_queries.rs
```diff
@@ -38,6 +38,16 @@ pub fn translate(env: &Env, state: State) -> Result<Vec<sql::ast::CommonTableExp
...
                                 Err(err) => Err(err.clone()),
                                 Ok(variables_table) => Ok(values::translate_variable(
+                                    // We need a 'State' value when translating variables in order
+                                    // to be able to generate fresh names for bound relational
+                                    // expressions.
+                                    // However, we cannot readily re-use the 'state' argument we're
+                                    // given, since that is an "owned" value which is "moved" in
+                                    // the course of iterating over the native queries accumulated
+                                    // within.
+                                    // Ideally we should resolve this by tracking native queries
```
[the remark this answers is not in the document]
plcplc: Not quite "just" pass it by reference.

Passing by reference results in a borrowing conflict, since we are looping over the native queries (also contained within the `state` value). Both the loop and `translate_variables` want to borrow the reference to the `state` value.

(From my understanding this prohibits `translate_variables` from mutating the list of native queries which the loop is iterating over, which makes sense IMO, as it _would_ be able to do just that with a `&mut State` reference.)

It's true that we could also clone our way out of this, but this also seems wasteful to me, and doesn't address the issue that we'd be passing somewhat nonsensical data to `translate_variables`, in the shape of redundant native queries that (according to the types) look like you could insert new ones into, but which would just be ignored.

PR #267, opened 2024-01-22
Unify code path for json value translation with variables

plcplc:
### What

Rather than having two different ways to translate values (depending on whether they are literals or variables) we instead use the same method in both cases.

### How

As a consequence we need to have the `State` available in literal value translation as well as in variable translation, which requires some follow-up changes in the various other translation functions.

We also duplicate the test of nested array types from the variables-case to the literal values-case.

### DIFF
--- a/crates/query-engine/translation/src/translation/query/native_queries.rs
+++ b/crates/query-engine/translation/src/translation/query/native_queries.rs
@@ -14,6 +14,17 @@ pub fn translate(env: &Env, state: State) -> Result<Vec<sql::ast::CommonTableExp
     let variables_table = env.get_variables_table();
     let native_queries = state.get_native_queries();
 
+    // We need a 'State' value when translating variables in order
+    // to be able to generate fresh names for bound relational
+    // expressions.
+    // However, we cannot readily re-use the 'state' argument we're
+    // given, since that is an "owned" value which is "moved" in
+    // the course of iterating over the native queries accumulated
+    // within.
+    // Ideally we should resolve this by tracking native queries
+    // separately from the fresh table name counter.
+    let mut translation_state = State::new();
+
     // for each found table expression
     for native_query in native_queries {
         // convert metadata representation to sql::ast representation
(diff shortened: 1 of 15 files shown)


crates/query-engine/translation/src/translation/query/native_queries.rs
```diff
@@ -33,7 +33,7 @@ pub fn translate(env: &Env, state: State) -> Result<Vec<sql::ast::CommonTableExp
                         None => Err(Error::ArgumentNotFound(param.clone())),
                         Some(argument) => match argument {
                             models::Argument::Literal { value } => {
-                                values::translate_json_value(value, &typ)
+                                values::translate_json_value(&mut State::new(), value, &typ)
```
SamirTalwar, line 36: Please document why we're constructing a new state here.

crates/query-engine/translation/tests/tests.rs
```diff
@@ -12,6 +12,12 @@ fn select_array_column() {
     insta::assert_snapshot!(result);
 }
 
+#[test]
+fn select_array_column_nested_types() {
+    let result = common::test_translation("select_array_column_nested_types").unwrap();
+    insta::assert_snapshot!(result);
+}
```
SamirTalwar, line 19: It's unclear to me; is this a test that would previously have failed, or is it a test that would have passed but was simply not present?

crates/query-engine/translation/src/translation/query/native_queries.rs
plcplc: Done~

crates/query-engine/translation/tests/tests.rs
plcplc: Sorry - that is indeed not clear.

This is a test that was just not present. It passed before and after this PR.

Each of the adding of the test and the revision of the values-translation-logic could sensibly have been submitted as separate PRs.

SamirTalwar (approved):

PR #271, opened 2024-01-30
Only get types and procedures from unqualified schemas

plcplc:
### What

In order to be able to unambiguously refer to types and procedures (i.e to avoid an issue like https://github.com/hasura/graphql-engine/issues/10100) we want to be sure that we track the schema types and procedures are defined in, similar to how we do tables.

This PR implements a minimal first step on the way to this, by simply only tracking types and procedures defined in schemas appearing in a new `unqualifiedSchemasForTypesAndProcedures` configuration option.

It is then up to a follow-up PR to implement schema name prefixing.

(Small aside: As an auxiliary improvement, thi PR introduces the use of the `regclass` etc. types to refer to catalog table entries instead of hardcoding oids. Incidentally this fixes fetching of table and column comments on Cockroach.)

### How

We add a new configuration option field `unqualifiedSchemasForTypesAndProcedures` that defaults to `["public","pg_catalog","tiger"]`. During introspection we only look in these schemas for types and procedures.

For consistency we also rename `unqualifiedSchemas` to `unqualifiedSchemasForTables`. In order to make this change conservative we still recognize `unqualifiedSchemas` in the API schema and json parsers, but we will only ever output `unqualifiedSchemasForTables` from the configuration server.

### DIFF
--- a/crates/connectors/ndc-postgres/src/configuration/version2.rs
+++ b/crates/connectors/ndc-postgres/src/configuration/version2.rs
@@ -50,14 +50,15 @@ impl RawConfiguration {
     }
 }
 
-pub fn default_unqualified_schemas() -> Vec<String> {
+pub fn default_unqualified_schemas_for_tables() -> Vec<String> {
+    vec!["public".to_string()]
+}
+
+pub fn default_unqualified_schemas_for_types_and_procedures() -> Vec<String> {
     vec![
-        // For the sake of having the user's tables to appear unqualified by default.
         "public".to_string(),
-        // For the sake of having types and procedures appear unqualified by default.
         "pg_catalog".to_string(),
         "tiger".to_string(),
-        // "topology".to_string(),
     ]
 }
 
@@ -71,8 +72,11 @@ pub struct ConfigureOptions {
     pub excluded_schemas: Vec<String>,
     /// The names of Tables and Views in these schemas will be returned unqualified.
     /// The default setting will set the `public` schema as unqualified.
-    #[serde(default = "default_unqualified_schemas")]
+    #[serde(default = "default_unqualified_schemas_for_tables")]
     pub unqualified_schemas: Vec<String>,
+    /// The types and procedures in these schemas will be returned unqualified.
+    #[serde(default = "default_unqualified_schemas_for_types_and_procedures")]
+    pub unqualified_schemas_for_types_and_procedures: Vec<String>,
     /// The mapping of comparison operator names to apply when updating the configuration
     #[serde(default = "default_comparison_operator_mapping")]
     pub comparison_operator_mapping: Vec<version1::ComparisonOperatorMapping>,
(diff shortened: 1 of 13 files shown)


crates/connectors/ndc-postgres/src/configuration/version2.rs
```diff
@@ -50,6 +50,17 @@ impl RawConfiguration {
     }
 }
 
+pub fn default_unqualified_schemas() -> Vec<String> {
```
soupi, line 53: I'm not sure I like this approach because it conflates things from different namespaces. Perhaps we need separate settings for unqualified schemas for each of the namespaces?

plcplc: I'm suspecting as much myself.

crates/connectors/ndc-postgres/src/configuration/version2.rs
```diff
@@ -50,6 +50,18 @@ impl RawConfiguration {
     }
 }
 
+pub fn default_unqualified_schemas_for_tables() -> Vec<String> {
```
soupi, line 53:
```suggestion
/// Collection names of tables in these schemas will be appear as unqualified.
pub fn default_unqualified_schemas_for_tables() -> Vec<String> {
```

crates/connectors/ndc-postgres/src/configuration/version2.rs
```diff
@@ -50,6 +50,18 @@ impl RawConfiguration {
...
+    vec!["public".to_string()]
+}
+
+pub fn default_unqualified_schemas_for_types_and_procedures() -> Vec<String> {
```
soupi, line 57:
```suggestion
/// Types, operators and procedures from these schemas will appear unqualified in the configuration.
pub fn default_unqualified_schemas_for_types_and_procedures() -> Vec<String> {
```

crates/connectors/ndc-postgres/src/configuration/version2.rs
```diff
@@ -58,10 +72,19 @@ pub struct ConfigureOptions {
     /// internal schemas of Postgres, Citus, Cockroach, and the PostGIS extension.
     #[serde(default = "default_excluded_schemas")]
     pub excluded_schemas: Vec<String>,
+    /// Deprecated alias for 'unqualifiedSchemasForTables'.
+    #[serde(default)]
+    // weirdly, we have to use 'skip_serializing_if' rather than 'skip_serializing', since the
+    // latter causes schemars to consider the field required, in spite of the 'default' attribute.
+    #[serde(skip_serializing_if = "always_skip_unqualified_schemas")]
+    pub unqualified_schemas: Vec<String>,
```
soupi, line 80: Can you please document the relationship between this field and the others? What happens if this field and the others are defined?

crates/connectors/ndc-postgres/src/configuration/version2.rs
```diff
@@ -93,6 +119,10 @@ impl Default for ConfigureOptions {
     }
 }
 
+pub fn always_skip_unqualified_schemas(_: &Vec<String>) -> bool {
```
soupi, line 122:
```suggestion
/// This is a deprecated field subsumed by `unqualified_schemas_for_tables` and
/// `unqualified_schemas_for_types_and_procedures`.
/// We don't want to output it when generating the configuration.
pub fn always_skip_unqualified_schemas(_: &Vec<String>) -> bool {
```

soupi (approved):

PR #272, opened 2024-01-31
CI: Merge the nix-build and deploy jobs.

SamirTalwar:
### What

We use `workflow_run` to trigger the "deploy" workflow after "nix build" has completed. However, this only seems to work on branches; pushing tags does _not_ trigger a job using `workflow_run`.

### How

We can work around this by merging the two jobs into a single workflow, and specifying the dependency chain there.


soupi:
is the nix build not used for other things, such as benchmarks?

SamirTalwar:
It is, and it will continue to build on all branches. It's only the deployment step that is skipped on non-`main` branches and tags.

Right now the "benchmarks" workflow doesn't rely on the "nix build" workflow. It just calls `nix build` itself, and relies on the cache to make sure it doesn't do much work most of the time. (Sometimes it will end up duplicating a build, which is probably fine.)

soupi (approved):

SamirTalwar:
Merging this; we will find out if it really works on `main`.

PR #273, opened 2024-01-31
Prepare the changelog for v0.3.0.

SamirTalwar:


changelog.md
```diff
@@ -2,6 +2,30 @@
...
+### Changed
+
+- Native query mutations are now marked as procedures instead of collections
+  ([#234](https://github.com/hasura/ndc-postgres/pull/234))
+- Introspection now avoids `DISTINCT ON` to be more compatible with PostgreSQL variants
+  ([#260](https://github.com/hasura/ndc-postgres/pull/260))
+
 ## [0.2.0] - 2023-12-21
 
 ### Added
```
SamirTalwar, line 31: We have two "Added" sections for v0.2.0. Not sure why. I will open a separate PR to change this.

changelog.md
```diff
@@ -2,6 +2,30 @@
...
 ## [Unreleased]
 
+## [0.3.0] - 2023-01-31
+
+### Added
+
+- Basic support for auto-generated insert procedures for tables.
+  ([#261](https://github.com/hasura/ndc-postgres/pull/261))
+- Support for composite types in inputs and outputs. Transformations are not yet
+  supported.
```
plcplc, line 12: 'Transformations'? Perhaps you meant:

```suggestion
- Support for composite types in inputs and outputs. Automatic introspection
  is not yet supported.
```

plcplc (approved):

changelog.md
soupi: The item listed under the top "Added" section of v0.2.0 are changed that are supposed to be in the changelog for v0.3.0

soupi: I probably added it in [this pr](https://github.com/hasura/ndc-postgres/commit/dd4172889830ecb5c6362e0be90b9c1106574363#diff-3bd14d078188074c410028847113ceae68865d0ad5b844a27183ef87fbe2fcc3) by mistake instead of above the v0.2.0 header.

changelog.md
SamirTalwar: I meant that we don't provide any operators on these composite types. I will add both.

```suggestion
- Support for composite types in inputs and outputs. Right now, no operators are
  provided for composite types, and there is no automatic introspection.
```

plcplc: Hm.. I still wonder if it's relevant to say that, since the ndc spec does not allow for any operators on things that are not scalar types.

Wording it like this suggests this is just something that has not been implemented in ndc-postgres yet.

changelog.md
SamirTalwar: Oh, you're right! I will update that. I see I actually duplicated some of them too.

changelog.md
SamirTalwar: I am personally OK with that. I think it's more important to state that we do or don't do something than _why_.

If someone cares, they will ask!

PR #274, opened 2024-01-31
Do not use `workflow_run` for end-to-end tests either.

SamirTalwar:
### What

These are no longer running because I changed the name of the workflow.

This upset me, as I would expect a warning, at least. Instead I have no feedback. This has convinced me that `workflow_run` is a bad idea.

### How

I am merging the E2E test job into the "ship" workflow so that we can use job chaining instead, which seems less brittle.


i-am-tom (approved):

SamirTalwar:
[Tested and working.](https://github.com/hasura/ndc-postgres/actions/runs/7729276785/job/21098910944)

PR #275, opened 2024-02-01
Revert the connector reference for postgres-config-server E2E tests.

SamirTalwar:
### What

I was a bit over-eager and overwrote this with the Docker tag. The config server tests do not need the Docker tag, they need the SHA, to clone the repository correctly.

### How

Reverted a line.


SamirTalwar:
[Tested; seems to work.](https://github.com/hasura/ndc-postgres/actions/runs/7740030774/job/21104283241)

soupi (approved):

PR #283, opened 2024-02-06
update ndc-spec to 0.1.0-rc14

soupi:
- depends on: https://github.com/hasura/ndc-hub/pull/93
- todo: add changelog entry

### What

We are updating connector to support the latest ndc-connector spec (0.1.0-rc14).
While fixing some breaking changes, such as renaming `where` to `predicate`, we also support new capabilities:

1. Explaining mutations via the `/mutation/explain` endpoint
2. Supporting filtering using `in` by columns and variables

### How

1. Since a mutation request may contain multiple mutation operations, the explain response will contain a field named `<operation_name> SQL Mutation` and a field named `<operation_name> Execution Plan` for each operation.
2. To filter in column or variable, we need to compare against a subquery of the shape `(select value from unnest(<column-or-variable>) as table(value))`.


crates/connectors/ndc-postgres/src/connector.rs
```diff
@@ -180,15 +179,41 @@ impl connector::Connector for Postgres {
...
+                err
+            })
+            .map(Into::into)
+    }
+
+    /// Explain a mutation by creating an execution plan
+    ///
+    /// This function implements the [mutation/explain endpoint](https://hasura.github.io/ndc-spec/specification/explain.html)
+    /// from the NDC specification.
+    async fn mutation_explain(
```
soupi, line 210: new endpoint and capability - explaining mutation requests.

crates/connectors/ndc-postgres/src/lib.rs
```diff
@@ -3,7 +3,6 @@
 pub mod capabilities;
 pub mod configuration;
 pub mod connector;
-pub mod explain;
```
soupi, line 6: replaced by a `query::explain` and `mutation::explain` modules.

crates/query-engine/sql/src/sql/ast.rs
```diff
@@ -108,6 +108,11 @@ pub enum From {
         alias: TableAlias,
         column: ColumnAlias,
     },
+    Unnest {
```
soupi, line 111: Used for fancy queries like `in <variable>` or `in <column>`, which are converted to a subquery of unnest.

crates/query-engine/translation/src/translation/query/filtering.rs
```diff
@@ -67,116 +68,153 @@ pub fn translate_expression(
             column,
             operator,
             value,
-        } => {
+        } if operator == "in" => {
```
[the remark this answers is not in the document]
soupi: the old binary array comparisons was only between a column and a vector of values. This new formulation lets us compare against scalar arrays, columns and variables. These are implemented below.

crates/query-engine/translation/src/translation/query/filtering.rs
```diff
@@ -67,116 +68,153 @@ pub fn translate_expression(
...
-            translate_exists_in_collection(
-                env,
-                state,
-                root_and_current_tables,
-                in_collection.clone(),
-                predicate,
-            )?,
-            vec![],
-        )),
+        } => match predicate {
```
soupi, line 205: predicates are now optional instead of being an equivalence of `true` when they are omitted.

crates/tests/databases-tests/src/postgres/snapshots/databases_tests__postgres__explain_tests__mutation__insert_artist_album.snap
soupi, line 1: explain mutations

crates/tests/tests-common/goldenfiles/select_where_in_variable.json
```diff
@@ -0,0 +1,48 @@
...
+      "type": "binary_comparison_operator",
+      "column": {
+        "type": "column",
+        "name": "three",
+        "path": []
+      },
+      "operator": "in",
+      "value": {
+        "type": "variable",
+        "name": "array"
```
soupi, line 21: new capability: filter by "in variable"

crates/query-engine/translation/src/translation/query/filtering.rs
```diff
@@ -368,27 +410,18 @@ fn translate_comparison_value(
     state: &mut State,
     root_and_current_tables: &RootAndCurrentTables,
     value: models::ComparisonValue,
-    typ: &database::ScalarType,
+    typ: database::Type,
```
SamirTalwar, line 413: These could probably both be some references to avoid clones at the call sites.

crates/query-engine/translation/tests/goldenfiles/sorting_by_nested_relationship_column_with_predicate/request.json
```diff
@@ -29,9 +29,7 @@
                         "name": "title",
                         "path": []
                       },
-                      "operator": {
-                        "type": "equal"
-                      },
+                      "operator": "eq",
```
SamirTalwar, line 32: Why is this one `"eq"` but others are `"_eq"`?

soupi: `eq` refers to the built-in equals operator from ndc-spec, `_eq` refers to the operator defined in the metadata of some tests (in tables.json). Most tests will use `eq`, but some tests might have wanted to test a metadata-defined operator? I wouldn't mind changing everything to `eq`.

crates/query-engine/translation/src/translation/query/filtering.rs
soupi: Done. Thanks

crates/query-engine/translation/tests/goldenfiles/sorting_by_nested_relationship_column_with_predicate/request.json
SamirTalwar: I honestly don't understand this well enough to agree or disagree, so I am going to pretend I didn't ask and trust that you know what you're doing.

soupi: ndc-spec defines two standard operators that a connector should implement - equal (`eq`) and in (`in`). Any other operator is a custom operator defined by the connector.

So for us we implement `in` and `eq` as built-in operators that do not need to be defined, and we also allow users to define their own operators in the ndc-metadata using the `comparisonOperators` field. [Example](https://github.com/hasura/ndc-postgres/blob/268e0ca0b68f9f41316dfa50746f563924f8d7bd/crates/query-engine/translation/tests/goldenfiles/select_where_unrelated_exists/tables.json#L46).

In most tests (like this one), we use the built-in `eq` operator. In other tests where we use the `_eq` operator, we have to define it like in the example I linked.

Tests that use `_eq` also test custom comparison operators, tests that use `eq` test the built-in operator.

SamirTalwar:
I think I made a big merge conflict for you. Sorry.

PR #285, opened 2024-02-12
Nix: Fix the build inputs for the shell and app.

SamirTalwar:
This makes a few changes to the Nix flake to get things building on my machine.

1. On macOS, add `CoreFoundation` to the list of dependencies.
2. Adds `strictDeps = true` to the package, which means we need to add `openssl.dev` (and `pkg-config`) to `nativeBuildInputs` as they're required at build-time.
3. Move `openssl` and `protobuf` to `buildInputs` as they're required at runtime.
4. Reuse inputs from the default package in the shell.
5. Hide the `perf` shell from macOS so `nix flake check` works.

The last point is partially undoing a change I made earlier, where the shell brought in inputs from all flake checks. This was way too broad; this narrows it a lot.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -83,13 +83,15 @@
         formatter = pkgs.nixpkgs-fmt;
 
         devShells.default = pkgs.mkShell {
+          # include dependencies of the default package
+          inputsFrom = [ self.packages.${localSystem}.default ];
+
           # build-time inputs
           nativeBuildInputs = [
             # Development
             pkgs.just
             pkgs.nixpkgs-fmt
             pkgs.nodePackages.prettier
-            pkgs.pkg-config
 
             # Rust
             pkgs.cargo-edit
@@ -107,12 +109,6 @@
 
             # Deployment
             pkgs.skopeo
-          ];
-
-          # runtime inputs
-          buildInputs = [
-            pkgs.openssl
-            pkgs.protobuf
           ];
         };
 
--- a/flake.nix
+++ b/flake.nix
@@ -82,49 +82,51 @@
 
         formatter = pkgs.nixpkgs-fmt;
 
-        devShells.default = pkgs.mkShell {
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -83,13 +83,15 @@
         formatter = pkgs.nixpkgs-fmt;
 
         devShells.default = pkgs.mkShell {
+          # include dependencies of the default package
+          inputsFrom = [ self.packages.${localSystem}.default ];
```
plcplc, line 87: .. Wasn't this the thing we _didn't_ want to do?

nix/app.nix
```diff
@@ -24,18 +24,18 @@ let
       in
       lib.cleanSourceWith { src = craneLib.path ./..; filter = isSourceFile; };
 
+    strictDeps = true;
```
plcplc, line 27: what's this do?

flake.nix
SamirTalwar: It was loading all the checks before. This is a bit more sensible.

nix/app.nix
SamirTalwar: It tells it not to get confused between `nativeBuildInputs` (build-time) and `buildInputs` (runtime).

flake.nix
plcplc: Ah, right. inputsFrom mentioned checks earlier, I remember

plcplc: So basically this time it doesn't include our crates themselves, but just their dependencies?

plcplc: (since our package itself was part of checks which was used previously)

plcplc (approved):

flake.nix
SamirTalwar: Exactly!

plcplc (approved):

PR #289, opened 2024-02-12
Pull out a configuration library.

SamirTalwar:
### What

We require a configuration library so that we can re-use configuration types and functions in both a CLI plugin _and_ a server-side connector build service.

This has been done with minimal effort.

Further work is required to minimize its surface area. Specifically, I believe we should:

1. reduce the number of crates depending on this one,
2. remove the dependency from "ndc-postgres-configuration" on "query-engine-sql" (it's only used to pull in `IsolationLevel`), and
3. refactor to reduce the number of exposed types and functions as much as possible.

I have not done any of these things in this changeset to make it easier to review, and because I didn't want reviews of this extraction to get bogged down in whether we should do the above or not.

### How

I moved `ndc-postgres/src/configuration.rs` and everything in `ndc-postgres/src/configuration` to their own crate, imaginatively named "ndc-postgres-configuration".

I then re-exported what was necessary from `lib.rs`, and changed imports in several places.


plcplc (approved):
LGTM

PR #290, opened 2024-02-12
Capture more/prettier information in translation snapshot tests.

SamirTalwar:
### What

Two things:

1. Always capture the pre- and post-query statements.
2. Pretty-print the query parameters, to make them more readable when they're large.

I think this makes the test snapshots a lot clearer.

### How

1. Remove a couple of `if` blocks.
2. Convert a vector of pairs to a `BTreeMap` (so it's sorted), and add a `#`.


soupi (approved):
Nice!

PR #291, opened 2024-02-12
Add a trait for `to_sql`.

SamirTalwar:
### What

This allows us to implement `to_sql` on types from a different package.

This is required to move `IsolationLevel` to a shared library, in order to stop `ndc-postgres-configuration` from depending on the entire `query-engine-sql` crate.

### How

We introduce the `ToSql` trait for all `fn to_sql` implementations. This means we need to import the trait in a few places.

This is only used in one place outside the `query-engine-sql` crate, in a snapshot test. I think this is probably a sign of a leaky abstraction, but I couldn't fix it quickly and it didn't seem like it should stop me from making this change.

### DIFF
--- a/crates/query-engine/sql/src/sql/convert.rs
+++ b/crates/query-engine/sql/src/sql/convert.rs
@@ -4,10 +4,12 @@ use super::ast::*;
 use super::helpers;
 use super::string::*;
 
-// Convert to SQL strings
+pub trait ToSql {
+    fn to_sql(&self, sql: &mut SQL);
+}
 
-impl With {
-    pub fn to_sql(&self, sql: &mut SQL) {
+impl ToSql for With {
+    fn to_sql(&self, sql: &mut SQL) {
         if self.common_table_expressions.is_empty() {
         } else {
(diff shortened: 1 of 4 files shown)


soupi:
Do we have any other alternative to adding a trait?
I prefer we only define traits that describe algebraic properties with laws. Traits are a very rigid construct that forces you to modify everything that share a trait in tandem. When there are no actual common properties these construct share, this translates to unnecessary inflexibility that will make modifying this code in the future more difficult. I feel that this is a very crucial part of the code and that this feature does not pay for it's weight. Perhaps we could think together of a different approach to solve the problem you are facing?

SamirTalwar:
@soupi: I can totally understand that.

We do have an alternative approach, which is to copy the `IsolationLevel` enum (it's small) to the configuration crate, and provide a very simple 1-to-1 mapping between the two independent-but-identical enums.

Do you think you'd prefer that?

soupi:
Yes I think that would be preferable because even if they are identical, they represent different things (one is configuration, the other is SQL ast) :) Thanks!

PR #292, opened 2024-02-13
Break dependencies through copying and pasting.

SamirTalwar:
### What

`ndc-postgres-configuration` had a dependency on `query-engine-sql` simply to use the `IsolationLevel` enum. This dependency is otherwise unnecessary and quite large.

### How

I have copied this enum into the configuration so that we can avoid this large dependency.

Of course, now we have two, so I've added a function inside the connector crate to convert from one to the other.

I have also improved the OpenAPI documentation on the `IsolationLevel` enum somewhat.

### DIFF
--- a/crates/connectors/ndc-postgres/src/configuration_mapping.rs
+++ b/crates/connectors/ndc-postgres/src/configuration_mapping.rs
@@ -1,3 +1,8 @@
+//! A module for converting ndc-configuration data type into query-engine data types.
+
+/// Convert a user-specified configuration of the isolation level for transactions
+/// into a SQL data type representing that isolation level, which will be passed
+/// to the engine on requests.
 pub(crate) fn convert_isolation_level(
     input: ndc_postgres_configuration::IsolationLevel,
 ) -> query_engine_sql::sql::ast::transaction::IsolationLevel {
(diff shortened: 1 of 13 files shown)


plcplc (approved):

crates/connectors/ndc-postgres/src/configuration_mapping.rs
```diff
@@ -0,0 +1,15 @@
+pub(crate) fn convert_isolation_level(
```
soupi, line 1: A docstring for the module and function would be nice :) Otherwise lgtm.

soupi, line 1: Maybe something like this?

```suggestion
//! A module for converting ndc-configuration data type into query-engine data types.

/// Convert a user-specified configuration of the isolation level for transactions
/// into a SQL data type representing that isolation level, which will be passed
/// to the engine on requests.
pub(crate) fn convert_isolation_level(
```

crates/tests/databases-tests/src/postgres/snapshots/databases_tests__postgres__configuration_tests__configuration_tests__get_rawconfiguration_v2_schema.snap
```diff
@@ -369,12 +369,29 @@ expression: schema
       }
     },
     "IsolationLevel": {
-      "description": "The isolation level of transactions",
-      "type": "string",
-      "enum": [
-        "ReadCommitted",
-        "RepeatableRead",
-        "Serializable"
+      "description": "The isolation level of the transaction in which a query is executed.",
```
soupi, line 372: Great!

crates/connectors/ndc-postgres/src/configuration_mapping.rs
SamirTalwar: I love it, thanks!

PR #293, opened 2024-02-13
Borrow more when constructing `RuntimeConfiguration`.

SamirTalwar:
### What

The documentation for `RuntimeConfiguration` states that it borrows whenever possible, which was false. I have made some changes to borrow (or copy, if copying is cheap) all fields.

### How

I introduced the `'request` lifetime and annotated fields with it. The metadata might be borrowed, or it might be converted from a v1 metadata, so this is wrapped in a `Cow` to abstract over this.

When possible, we now pass around a `RuntimeConfiguration` by value rather than by reference.

I also added `Copy` to `MutationsVersion` so that we can freely pass it around without having to worry about borrowing. `IsolationLevel` already implements `Copy`.

### DIFF
--- a/crates/configuration/src/configuration.rs
+++ b/crates/configuration/src/configuration.rs
@@ -1,5 +1,7 @@
 //! Configuration for the connector.
 
+use std::borrow::Cow;
+
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
@@ -76,7 +78,7 @@ pub async fn validate_raw_configuration(
 /// borrowed rather than owned.
 #[derive(Debug)]
 pub struct RuntimeConfiguration<'request> {
-    pub metadata: metadata::Metadata,
+    pub metadata: Cow<'request, metadata::Metadata>,
     pub pool_settings: &'request version2::PoolSettings,
     pub connection_uri: &'request str,
     pub isolation_level: version2::IsolationLevel,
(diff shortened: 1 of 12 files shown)


SamirTalwar:
Seems to get a reasonable speedup of around 10% in the benchmarks.

crates/configuration/src/configuration.rs
```diff
@@ -1,5 +1,7 @@
 //! Configuration for the connector.
 
+use std::borrow::Cow;
```
plcplc, line 3: #TIL about Cow. This is very neat.

crates/configuration/src/configuration.rs
```diff
@@ -75,34 +77,34 @@ pub async fn validate_raw_configuration(
...
+    pub metadata: Cow<'request, metadata::Metadata>,
+    pub pool_settings: &'request version2::PoolSettings,
+    pub connection_uri: &'request str,
     pub isolation_level: version2::IsolationLevel,
     pub mutations_version: Option<metadata::mutations::MutationsVersion>,
 }
 
 /// Apply the common interpretations on the Configuration API type into an RuntimeConfiguration.
-pub fn as_runtime_configuration(config: &Configuration) -> RuntimeConfiguration {
+pub fn as_runtime_configuration(config: &Configuration) -> RuntimeConfiguration<'_> {
```
plcplc, line 89: I'm not opposed to us using Cow per se, but could we have achieved the same (i.e., avoiding cloning) with ownership:

```
pub fn as_runtime_configuration(config: Configuration) -> RuntimeConfiguration {
```

.. supposing the caller of this function doesn't need to use its `config` value again. I don't remember if it does though.

SamirTalwar: Unfortunately we get a reference, and this is baked into ndc-hub.

plcplc: Oh well, in that case, bring in the COWs

plcplc (approved):

PR #294, opened 2024-02-13
CI: Cap the benchmark history in charts to 50 items.

SamirTalwar:
### What

Each data point on the benchmark charts is so close to the next one it's hard to hover over.

In addition, a drastic change isn't even really visible until it's been a few commits.

### How

We cap the maximum items in each chart to 50.


SamirTalwar:
This might be too big or too small. I suggest we merge it and find out.

soupi (approved):

PR #295, opened 2024-02-13
Replace configuration with a new version, identical to v2.

SamirTalwar:
### What

This introduces configuration version 3, and removes versions 1 and 2.

This is most definitely a breaking change, and we're not done. Future changes will tear this apart and rebuild it, but I wanted to make this part easier to review. 

### How

I did this in a few steps to make sure I didn't break anything.

1. I pulled out some of the lower-level configuration types to their own files.
2. Next, I copied the remaining shared code from v1 to v2 so there were no cross-dependencies.
3. Then I copied v2 to v3, and modified all the tests to use a v3 config (identical to v2 except the `"version"` field).
4. Finally, I removed v1 and v2.


crates/configuration/src/version3.rs
```diff
@@ -118,12 +91,12 @@ pub struct ConfigureOptions {
 impl Default for ConfigureOptions {
     fn default() -> ConfigureOptions {
         ConfigureOptions {
-            excluded_schemas: version1::default_excluded_schemas(),
+            excluded_schemas: default_excluded_schemas(),
             unqualified_schemas: vec![],
```
plcplc, line 95: Another desired change for version 3 is to remove `unqualified_schemas`. This can happen in this PR or a follow-up.

plcplc: And in the same vein it'd be good to rename `excluded_schemas` to `excluded_schemas_for_tables` or similar.

justfile
```diff
@@ -223,29 +215,20 @@ test *args: start-dependencies create-aurora-ndc-metadata
...
-  ./scripts/archive-old-ndc-metadata.sh '{{POSTGRES_V1_CHINOOK_NDC_METADATA}}'
-  ./scripts/generate-chinook-configuration.sh 'ndc-postgres' '{{POSTGRESQL_CONNECTION_STRING}}' '{{POSTGRES_V1_CHINOOK_NDC_METADATA}}'
-  ./scripts/generate-chinook-configuration.sh 'ndc-postgres' '{{CITUS_CONNECTION_STRING}}' '{{CITUS_V1_CHINOOK_NDC_METADATA}}'
-  ./scripts/generate-chinook-configuration.sh 'ndc-postgres' '{{COCKROACH_CONNECTION_STRING}}' '{{COCKROACH_V1_CHINOOK_NDC_METADATA}}'
-
-  ./scripts/archive-old-ndc-metadata.sh '{{POSTGRES_V2_CHINOOK_NDC_METADATA}}'
-  ./scripts/generate-chinook-configuration.sh 'ndc-postgres' '{{POSTGRESQL_CONNECTION_STRING}}' '{{POSTGRES_V2_CHINOOK_NDC_METADATA}}'
-  ./scripts/generate-chinook-configuration.sh 'ndc-postgres' '{{CITUS_CONNECTION_STRING}}' '{{CITUS_V2_CHINOOK_NDC_METADATA}}'
-  ./scripts/generate-chinook-configuration.sh 'ndc-postgres' '{{COCKROACH_CONNECTION_STRING}}' '{{COCKROACH_V2_CHINOOK_NDC_METADATA}}'
+  ./scripts/archive-old-ndc-metadata.sh '{{POSTGRES_V3_CHINOOK_NDC_METADATA}}'
```
plcplc, line 218: It'd be good to leave some sort of marker that this is a place which should do something for all currently supported versions and not just the lastest.

plcplc: This is an issue in general. When we reduce our set of supported versions to just the latest it becomes harder to add a new version to that set, because you have to rediscover the parts that are different between the latest version and the full set of supported legacy versions.

plcplc (approved):
I'm good with the functional contents of this PR, but it would be nice if we could mark up the places that need to concern themselves with legacy supported versions when we bump the version.

crates/configuration/src/version3.rs
SamirTalwar: Yup, I already made the change but wanted to keep this diff smaller.

justfile
SamirTalwar: We will probably have to reconsider the idea of a "version" as soon as we split the configuration into multiple files.

plcplc: Even then, knowing which parts of the supporting dev code historically had to take older versions into account is probably a good baseline understanding IMO.

SamirTalwar:
Gonna merge this on the grounds that I don't know what "versioning" looks like in the future. Let's track it.

PR #296, opened 2024-02-14
Bring ndc-postgres in line with ndc-spec RC15

i-am-tom:
## What

This PR updates `ndc-postgres` to adhere to the RC15 version of `ndc-spec`.

## How

This required a bit more work than we'd originally expected: because the format of procedure results changed, we now need to generate a new object type for each procedure (namely, the result type: a product of `affected_rows` (int4) and `returning` (whatever the underlying collection's row type is).

### DIFF
--- a/crates/configuration/Cargo.toml
+++ b/crates/configuration/Cargo.toml
@@ -6,7 +6,7 @@ edition = "2021"
 [dependencies]
 query-engine-metadata = { path = "../query-engine/metadata" }
 
-ndc-sdk = { git = "https://github.com/hasura/ndc-hub.git", rev = "fc27ac5" }
+ndc-sdk = { git = "https://github.com/hasura/ndc-hub.git", rev = "e0e9629" }
 
 schemars = { version = "0.8.16", features = ["smol_str", "preserve_order"] }
 serde = "1.0.196"
--- a/crates/configuration/Cargo.toml
+++ b/crates/configuration/Cargo.toml
@@ -6,7 +6,7 @@ edition = "2021"
 [dependencies]
 query-engine-metadata = { path = "../query-engine/metadata" }
 
-ndc-sdk = { git = "https://github.com/hasura/ndc-hub.git", rev = "02d26c1" }
+ndc-sdk = { git = "https://github.com/hasura/ndc-hub.git", rev = "fc27ac5" }
 
 schemars = { version = "0.8.16", features = ["smol_str", "preserve_order"] }
 serde = "1.0.196"
--- a/crates/connectors/ndc-postgres/Cargo.toml
+++ b/crates/connectors/ndc-postgres/Cargo.toml
@@ -21,7 +21,7 @@ query-engine-metadata = { path = "../../query-engine/metadata" }
 query-engine-sql = { path = "../../query-engine/sql" }
 query-engine-translation = { path = "../../query-engine/translation" }
 
-ndc-sdk = { git = "https://github.com/hasura/ndc-hub.git", rev = "fc27ac5" }
+ndc-sdk = { git = "https://github.com/hasura/ndc-hub.git", rev = "e0e9629" }
 
 async-trait = "0.1.77"
 percent-encoding = "2.3.1"
--- a/crates/connectors/ndc-postgres/Cargo.toml
+++ b/crates/connectors/ndc-postgres/Cargo.toml
@@ -21,7 +21,7 @@ query-engine-metadata = { path = "../../query-engine/metadata" }
 query-engine-sql = { path = "../../query-engine/sql" }
 query-engine-translation = { path = "../../query-engine/translation" }
 
-ndc-sdk = { git = "https://github.com/hasura/ndc-hub.git", rev = "02d26c1" }
+ndc-sdk = { git = "https://github.com/hasura/ndc-hub.git", rev = "fc27ac5" }
 
 async-trait = "0.1.77"
 percent-encoding = "2.3.1"
@@ -34,8 +34,8 @@ tracing = "0.1.40"
 url = "2.5.0"
 
 [dev-dependencies]
-ndc-client = { git = "https://github.com/hasura/ndc-spec.git", tag = "v0.1.0-rc.14" }
-ndc-test = { git = "https://github.com/hasura/ndc-spec.git", tag = "v0.1.0-rc.14" }
+ndc-client = { git = "https://github.com/hasura/ndc-spec.git", tag = "v0.1.0-rc.15" }
+ndc-test = { git = "https://github.com/hasura/ndc-spec.git", tag = "v0.1.0-rc.15" }
 tests-common = { path = "../../tests/tests-common" }
 
 axum = "0.6.20"
--- a/crates/connectors/ndc-postgres/src/schema.rs
+++ b/crates/connectors/ndc-postgres/src/schema.rs
@@ -437,6 +437,11 @@ fn make_procedure_type(
     let mut fields = BTreeMap::new();
     let object_type_name = format!("{name}_response");
 
+    // If int4 doesn't exist anywhere else in the schema, we need to add it here. However, a user
+    // can't filter or aggregate based on the affected rows of a procedure, so we don't need to add
+    // any aggregate functions or comparison operators. However, if int4 exists elsewhere in the
+    // schema and has already been added, it will also already contain these functions and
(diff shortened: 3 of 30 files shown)


crates/configuration/Cargo.toml
```diff
@@ -6,7 +6,7 @@ edition = "2021"
 [dependencies]
 query-engine-metadata = { path = "../query-engine/metadata" }
 
-ndc-sdk = { git = "https://github.com/hasura/ndc-hub.git", rev = "02d26c1" }
+ndc-sdk = { git = "https://github.com/hasura/ndc-hub.git", rev = "fc27ac5" }
```
soupi, line 9: We'll need to release this in ndc-hub before merging.

crates/connectors/ndc-postgres/src/schema.rs
```diff
@@ -393,12 +406,83 @@ fn insert_to_procedure(
...
+    // can't filter or aggregate based on the affected rows of a procedure, so we don't need to add
+    // any aggregate functions or comparison operators. However, if int4 exists elsewhere in the
+    // schema and has already been added, it will also already contain these functions and
+    // operators.
+    scalar_types
+        .entry("int4".to_string())
+        .or_insert(models::ScalarType {
+            aggregate_functions: BTreeMap::new(),
+            comparison_operators: BTreeMap::new(),
+        });
```
SamirTalwar, line 450: Is this also required when generating the configuration?

If it is, then it will need to be moved to `occurring_scalar_types`.

i-am-tom: @soupi

SamirTalwar: Thinking about it, even if it's not, that still feels like a better place for it. I plan on making that function private and putting the scalar types into the configuration at some point soon, so this function can just read them directly.

soupi: It's not really part of the configuration. Even if the user omits `int4` from the configuration we still need to expose this type for the schema or else procedures will break, so I don't see a way out of defining it here.

SamirTalwar: I am convinced.

Any reason not to add it in up-front rather than mutating later?

i-am-tom: Only that, if no mutations are generated, it's a type without a purpose, which we seem to be avoiding everywhere else? idk, I'm not sure what the danger is in having all scalar types declared upfront and not having the `ocurring_scalar_types` check, but doing it this way means we're still respecting that approach. Probably a wider question

crates/configuration/Cargo.toml
```diff
@@ -6,7 +6,7 @@ edition = "2021"
 [dependencies]
 query-engine-metadata = { path = "../query-engine/metadata" }
 
-ndc-sdk = { git = "https://github.com/hasura/ndc-hub.git", rev = "02d26c1" }
+ndc-sdk = { git = "https://github.com/hasura/ndc-hub.git", rev = "e0e9629" }
```
SamirTalwar, line 9: We can update to RC 16 directly if you like.

```suggestion
ndc-sdk = { git = "https://github.com/hasura/ndc-hub.git", rev = "ee52bae" }
```

crates/connectors/ndc-postgres/Cargo.toml
```diff
@@ -21,7 +21,7 @@ query-engine-metadata = { path = "../../query-engine/metadata" }
 query-engine-sql = { path = "../../query-engine/sql" }
 query-engine-translation = { path = "../../query-engine/translation" }
 
-ndc-sdk = { git = "https://github.com/hasura/ndc-hub.git", rev = "02d26c1" }
+ndc-sdk = { git = "https://github.com/hasura/ndc-hub.git", rev = "e0e9629" }
```
SamirTalwar, line 24:
```suggestion
ndc-sdk = { git = "https://github.com/hasura/ndc-hub.git", rev = "ee52bae" }
```

crates/connectors/ndc-postgres/Cargo.toml
```diff
@@ -34,8 +34,8 @@ tracing = "0.1.40"
 url = "2.5.0"
 
 [dev-dependencies]
-ndc-client = { git = "https://github.com/hasura/ndc-spec.git", tag = "v0.1.0-rc.14" }
-ndc-test = { git = "https://github.com/hasura/ndc-spec.git", tag = "v0.1.0-rc.14" }
+ndc-client = { git = "https://github.com/hasura/ndc-spec.git", tag = "v0.1.0-rc.15" }
+ndc-test = { git = "https://github.com/hasura/ndc-spec.git", tag = "v0.1.0-rc.15" }
```
SamirTalwar, line 38:
```suggestion
ndc-client = { git = "https://github.com/hasura/ndc-spec.git", tag = "v0.1.0-rc.16" }
ndc-test = { git = "https://github.com/hasura/ndc-spec.git", tag = "v0.1.0-rc.16" }
```

crates/query-engine/translation/Cargo.toml
```diff
@@ -5,7 +5,7 @@ edition = "2021"
 license = "Apache-2.0"
 
 [dependencies]
-ndc-sdk = { git = "https://github.com/hasura/ndc-hub.git", rev = "02d26c1" }
+ndc-sdk = { git = "https://github.com/hasura/ndc-hub.git", rev = "e0e9629" }
```
SamirTalwar, line 8:
```suggestion
ndc-sdk = { git = "https://github.com/hasura/ndc-hub.git", rev = "ee52bae" }
```

crates/tests/tests-common/Cargo.toml
```diff
@@ -9,9 +9,9 @@ name = "tests_common"
 path = "src/lib.rs"
 
 [dependencies]
-ndc-client = { git = "https://github.com/hasura/ndc-spec.git", tag = "v0.1.0-rc.14" }
-ndc-sdk = { git = "https://github.com/hasura/ndc-hub.git", rev = "02d26c1" }
-ndc-test = { git = "https://github.com/hasura/ndc-spec.git", tag = "v0.1.0-rc.14" }
+ndc-client = { git = "https://github.com/hasura/ndc-spec.git", tag = "v0.1.0-rc.15" }
+ndc-sdk = { git = "https://github.com/hasura/ndc-hub.git", rev = "e0e9629" }
+ndc-test = { git = "https://github.com/hasura/ndc-spec.git", tag = "v0.1.0-rc.15" }
```
SamirTalwar, line 14:
```suggestion
ndc-client = { git = "https://github.com/hasura/ndc-spec.git", tag = "v0.1.0-rc.16" }
ndc-sdk = { git = "https://github.com/hasura/ndc-hub.git", rev = "ee52bae" }
ndc-test = { git = "https://github.com/hasura/ndc-spec.git", tag = "v0.1.0-rc.16" }
```

crates/tests/databases-tests/Cargo.toml
```diff
@@ -23,7 +23,7 @@ postgres = []
 openapi-generator = { path = "../../documentation/openapi" }
 ndc-postgres = { path = "../../connectors/ndc-postgres" }
 ndc-postgres-configuration = { path = "../../configuration" }
-ndc-test = { git = "https://github.com/hasura/ndc-spec.git", tag = "v0.1.0-rc.14" }
+ndc-test = { git = "https://github.com/hasura/ndc-spec.git", tag = "v0.1.0-rc.15" }
```
SamirTalwar, line 26:
```suggestion
ndc-test = { git = "https://github.com/hasura/ndc-spec.git", tag = "v0.1.0-rc.16" }
```

SamirTalwar (commented):
If you want to go straight to ndc-spec v0.1.0-rc.16, here's the places you need to hit.

I suggest not accepting these comments because you'll also need to regenerate `Cargo.lock`.

SamirTalwar (approved):

crates/connectors/ndc-postgres/src/schema.rs
soupi: I think it's better to keep the cause and effect together. We generate the type because the procedure needs it, and we don't if we don't. It's better to place them together instead of in two separate places.

soupi:
I think separate steps are better.

PR #297, opened 2024-02-14
Reduce the surface area of the configuration crate.

SamirTalwar:
### What

This makes a few changes to the configuration (version 3) to reduce the surface area.

All changes are refactors _except_ for adding scalar types to the metadata.

### How

1. `occurring_scalar_types` is now private; we store the results in the metadata instead.
2. `unqualified_schemas` was deprecated and has been removed.
3. `set_connection_uri` is only used in tests, so I moved it to the tests.
4. `ConfigureOptions` has been moved to its own file to keep things a bit more private.
5. Other functions have been made private.
6. When configuring, we avoid cloning where possible.


justfile
```diff
@@ -234,6 +234,8 @@ generate-chinook-configuration: build start-dependencies
     echo "$(tput bold)$(tput setaf 3)WARNING:$(tput sgr0) Not updating the Aurora configuration because the connection string is unset."; \
   fi
 
+  ./scripts/archive-old-ndc-metadata.sh '{{POSTGRES_V3_CHINOOK_NDC_METADATA}}'
```
plcplc, line 237: I don't think there's much point to keeping these older versions around in this breaking-changes-epoch. (Even this PR makes changes to an archived version, which is non-sensical)

crates/query-engine/metadata/src/metadata/mod.rs
```diff
@@ -15,6 +15,8 @@ use serde::{Deserialize, Serialize};
 #[derive(Clone, Debug, Default, Serialize, Deserialize, JsonSchema)]
 #[serde(rename_all = "camelCase")]
 pub struct Metadata {
+    #[serde(default)]
```
plcplc, line 18: I don't think `scalar_types` belong in `Metadata` because the set of scalar types is not itself a root observable/introspected entity but instead derived from the column types and comparison operators etc.

I don't think having the set of scalar types be part of the metadata enables any new behaviors for a user, except maybe the ability to provoke the ndc to expose an inconsistent schema (I'm not sure).

IMO it'd be more appropriate to have the scalar types be part of `RuntimeConfiguration` instead.

plcplc (commented):
Some thoughts, but otherwise nice.

justfile
SamirTalwar: Sure, I'll just remove it.

crates/query-engine/metadata/src/metadata/mod.rs
SamirTalwar: OK, let's revert this for now and bring it back once `RuntimeConfiguration` is stored rather than re-computed per request.

plcplc (approved):

PR #298, opened 2024-02-15
update rust toolchain

soupi:
### What

We updated the rust toolchain for ndc-hub to 1.76.0, might as well do the same for ndc-postgres.


soupi:
@SamirTalwar I don't know how to update nix.

SamirTalwar:
@soupi: I think we just need to update the Rust overlay. I'll take care of it and push here, if it's OK with you.

SamirTalwar (approved):

PR #301, opened 2024-02-16
benchmarks: Generate configuration without the configuration server.

SamirTalwar:
### What

The configuration server is going away. Eventually we'll have a CLI that does the same thing, but that doesn't currently exist.

Fortunately, specifically for the benchmarks, we can replicate the configuration server behavior simply by using `jq` to replace the connection URI in the already-generated configuration file.

### How

I've made the changes in both the local benchmark script and in the CI YAML file.


i-am-tom (approved):
👍

PR #305, opened 2024-02-20
Switch to a configuration directory.

SamirTalwar:
### What

This upgrades ndc-sdk, which includes two major changes:

1. Configuration is now a directory, not a `RawConfiguration`.
2. The configuration server has gone.

### How

This means we have to make a lot of changes. I'll try and summarize them.

1. We now expect the configuration file to be a file inside the configuration directory. Test data and tests have been changed accordingly.
2. The configuration directory is assumed to be `/etc/connector` unless otherwise specified when run as a Docker image.
3. The configuration server has been removed. This means that our scripts that generate configuration no longer work. We will need to replace that logic with a CLI or similar.
4. We don't need to deploy the configuration server any more, because it doesn't exist.


justfile
```diff
@@ -213,27 +157,28 @@ test *args: start-dependencies create-aurora-ndc-metadata
...
-    echo "$(tput bold)$(tput setaf 3)WARNING:$(tput sgr0) Not updating the Yugabyte configuration because we are running on a non-x86_64 architecture."; \
-  fi
-  @ if [[ -n '{{AURORA_CONNECTION_STRING}}' ]]; then \
-    echo "$(tput bold)./scripts/generate-chinook-configuration.sh 'ndc-postgres' '{{AURORA_CONNECTION_STRING}}' '{{AURORA_V3_CHINOOK_NDC_METADATA_TEMPLATE}}'$(tput sgr0)"; \
-    ./scripts/generate-chinook-configuration.sh "ndc-postgres" '{{AURORA_CONNECTION_STRING}}' '{{AURORA_V3_CHINOOK_NDC_METADATA_TEMPLATE}}'; \
-    just create-aurora-ndc-metadata; \
-  else \
-    echo "$(tput bold)$(tput setaf 3)WARNING:$(tput sgr0) Not updating the Aurora configuration because the connection string is unset."; \
-  fi
+# THIS DOES NOT WORK.
```
soupi, line 160: I'm weary about removing this without an alternative because I used this yesterday.

SamirTalwar: We need to make this change to make the CLI work, and we need to make the CLI work to replace this.

I shall endeavour to replace it as urgently as possible.

soupi (approved):
Let's go

PR #306, opened 2024-02-20
Use the `RuntimeConfiguration` as the real configuration.

SamirTalwar:
### What

As we no longer need to persist this, we don't need to worry about the schema changing at all.

### How

I renamed `RuntimeConfiguration` to `Configuration`, removed the borrowing (making it a totally owned type), and fixed all compilation failures.

I have upgraded `ndc-sdk` to remove the expectation that `Configuration` must be serializable.


soupi (approved):
lgtm!

PR #307, opened 2024-02-20
Create a CLI which will initialize a directory with a config file.

SamirTalwar:
### What

This will initialize the current directory with a single file, `configuration.json`, with an empty configuration.

If the directory contains any files, the command will refuse to run, as a precaution.

### How

It wraps the `configuration::empty` function.

### DIFF
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -1,5 +1,4 @@
 use std::fs;
-use std::io;
 use std::path::Path;
 
 use clap::{Parser, Subcommand};
@@ -14,8 +13,14 @@ pub struct Args {
 
 #[derive(Debug, Clone, Subcommand)]
 pub enum Command {
-    /// Initialize a configuration in the current directory.
+    /// Initialize a configuration in the current (empty) directory.
     Initialize,
+}
+
+#[derive(Debug, PartialEq, thiserror::Error)]
+pub enum Error {
+    #[error("directory is not empty")]
+    DirectoryIsNotEmpty,
 }
 
 pub fn run(command: Command, configuration_dir: &Path) -> anyhow::Result<()> {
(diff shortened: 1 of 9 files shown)


crates/cli/src/lib.rs
```diff
@@ -0,0 +1,50 @@
+use std::fs;
```
soupi, line 1: please docstrings on modules types and functions they really help me more than you think.

SamirTalwar: Oh, thank you for the reminder! I will add lots of comments.

SamirTalwar: Done.

crates/cli/src/lib.rs
```diff
@@ -0,0 +1,62 @@
+//! The interpretation of the commands that the CLI cah handle.
+//!
+//! The CLI can do a few things. This provides a central point where those things are routed and
+//! then done, making it easier to test this crate deterministically.
```
soupi, line 4:
```suggestion
//! The interpretation of the commands that the CLI can handle.
//!
//! The CLI can do a few things. This provides a central point where those things are routed and
//! then done, making it easier to test this crate deterministically.
```

What is the purpose of the CLI? Is it only to initialize the config?

plcplc (approved):
And we're off to the races!

crates/cli/src/lib.rs
SamirTalwar: Yes, for now. I added a bit more commentary to `main.rs`.

crates/cli/src/lib.rs
```diff
@@ -0,0 +1,62 @@
...
+
+/// Run a command in a given directory.
+pub fn run(command: Command, configuration_dir: &Path) -> anyhow::Result<()> {
+    match command {
+        Command::Initialize => initialize(configuration_dir)?,
+    };
+    Ok(())
+}
+
+/// Initialize an empty directory with the connector configuration.
```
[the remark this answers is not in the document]
SamirTalwar: Stolen; thanks!

PR #311, opened 2024-02-21
Print useful output when ndc-test fails.

SamirTalwar:
### What

The `Debug` output of `Vec<FailedTest>` can be very long and hard to read.

### How

By simply pretty-printing each failure separated by newlines, we can make it readable.


soupi (approved):
Cool thanks!

PR #312, opened 2024-02-22
Add an `update` command to the CLI that introspects the database.

SamirTalwar:
### What

This works similarly to the update behavior of the old configuration server.

### How

There wasn't a lot to do.

This does not update the scripts; I will follow up with that later.

### DIFF
--- a/crates/cli/tests/update_tests.rs
+++ b/crates/cli/tests/update_tests.rs
@@ -1,0 +1,42 @@
+use std::fs;
+
+use ndc_postgres_cli::*;
+use ndc_postgres_configuration as configuration;
+use ndc_postgres_configuration::RawConfiguration;
+
+const CONNECTION_STRING: &str = "postgresql://postgres:password@localhost:64002";
+
+#[tokio::test]
+async fn test_update_configuration() -> anyhow::Result<()> {
+    let dir = tempfile::tempdir()?;
+    {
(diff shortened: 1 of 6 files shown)


crates/cli/tests/update_tests.rs
```diff
@@ -0,0 +1,42 @@
+use std::fs;
+
+use ndc_postgres_cli::*;
+use ndc_postgres_configuration as configuration;
+use ndc_postgres_configuration::RawConfiguration;
+
+const CONNECTION_STRING: &str = "postgresql://postgres:password@localhost:64002";
+
+#[tokio::test]
+async fn test_update_configuration() -> anyhow::Result<()> {
```
soupi, line 10: When does this run? As part of `just dev` and `just test`?

soupi (approved):
LGTM. You mentioned that you can run this with `cargo run --bin ndc-postgres-cli -- <args>`. Would be nice if as a follow-up this could be added as a `just` command and the development.md doc could be updated accordingly as well.

crates/cli/tests/update_tests.rs
SamirTalwar: Yeah, it's a normal test case; connects to the database just like the query tests do.

SamirTalwar:
@soupi: I'm gonna update the scripts right now, I just wanted to do it in a separate PR to make review easier. 😁

PR #314, opened 2024-02-22
Remove duplication in error types and error-handling.

SamirTalwar:
### What

I started trying to re-use the transaction handling logic we have around mutations for queries too, but got stuck because I either needed to duplicate a lot of code, which felt suspect and error-prone, or deduplicate a lot of other code first.

I chose the latter.

This deduplicates the error types between queries and mutations in a new `query_engine_execution::error::Error` type.

I have also deduplicated the use of this type and the `query_engine_translation::translation::error::Error` type, particularly when it comes to (a) recording the error for tracing and metrics, and (b) converting the error to an ndc-sdk `connector::QueryError`.

I think the separation of concerns in the resulting code makes it easier to read (large `.map_err` blocks can be something of a handful). This will also make it much easier to share transaction code between queries and mutations.

### How

First, I made the new error type. This meant I had to handle some extra cases in certain places, and I didn't like adding the code, so I decided instead to remove it.

I created two new files:

1. `record.rs`, which records errors with the tracing and metrics services.
2. `convert.rs, which converts errors.

From there it was a fairly mechanical process of using those new functions instead of the inline recording and conversion from before.

I removed the `Multiple` error case, instead deciding to just throw away the second error from rolling back in favor of the first one, which is more likely to contain useful information.

I also removed a bit of code:

* `ndc_postgres::query::plan_query` was duplicated in `ndc_postgres::query::explain::plan_query`, with different error-handling. By moving error-handling up a layer, the discrepancy went away and I was able to reuse the first one.
* `ndc_postgres::mutation::explain` had an additional error conversion function which converted the ndc-sdk error _back_ to an internal error. Again, by moving error-handling up a layer, this was no longer necessary.

### DIFF
--- a/crates/connectors/ndc-postgres/src/convert.rs
+++ b/crates/connectors/ndc-postgres/src/convert.rs
@@ -1,0 +1,75 @@
+//! Functions to convert between internal error types and the error types exposed by ndc-sdk.
+
+use ndc_sdk::connector;
+
+pub(crate) fn execution_error_to_query_error(
+    error: query_engine_execution::error::Error,
+) -> connector::QueryError {
+    match error {
+        query_engine_execution::error::Error::Query(query_error) => match &query_error {
+            query_engine_execution::error::QueryError::VariableNotFound(_) => {
+                connector::QueryError::InvalidRequest(query_error.to_string())
+            }
+            query_engine_execution::error::QueryError::NotSupported(_) => {
+                connector::QueryError::UnsupportedOperation(query_error.to_string())
+            }
+            query_engine_execution::error::QueryError::DBError(_) => {
+                connector::QueryError::UnprocessableContent(query_error.to_string())
+            }
+            query_engine_execution::error::QueryError::DBConstraintError(_) => {
+                connector::QueryError::UnprocessableContent(query_error.to_string())
+            }
+        },
+        query_engine_execution::error::Error::DB(_) => {
+            connector::QueryError::Other(Box::new(error))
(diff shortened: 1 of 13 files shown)


crates/connectors/ndc-postgres/src/convert.rs
```diff
@@ -0,0 +1,117 @@
+//! Functions to convert between internal error types and the error types exposed by ndc-sdk.
+
+use ndc_sdk::connector;
+
+pub(crate) fn execution_error_to_query_error(
```
soupi, line 5: Why pub(crate) and not just pub? I'm assuming that it could be useful to use these functions outside this crate, and there's no harm in that either. I think we can default to simple pub.

soupi (commented):
Looks good! One this that I would like is if we were to move `record.rs` and `convert.rs` under a common directory (named `error/` I guess) and have:

```
error/
  error/mod.rs
  error/convert.rs
  error/record.rs
```

They feel very specific and probably do not belong at the top-level src dir.

crates/connectors/ndc-postgres/src/convert.rs
soupi: And could we please add docstrings to these functions as well? :)

SamirTalwar: All very helpful suggestions. Thanks! I have done them all.

soupi (approved):
Thanks!

PR #319, opened 2024-02-23
Remove the OpenAPI and JSON Schema configuration schemas.

SamirTalwar:
### What

The configuration is no longer a single JSON file, and it therefore makes no sense for it to still have a schema.

It's time to remove it, sad though it is to see the work discarded.

### How

image


SamirTalwar:
Closing because we have better ideas.

PR #325, opened 2024-02-26
Read the connection URI from the environment.

SamirTalwar:
### What

The connection URI can now be provided from the environment, using `"uri": {"variable": "..."}`.

This means that we can use the same configuration in development and production, and enables the CLI or Cloud build infrastructure to fill it in with e.g. a region-specific URL.

### How

We replace the `ResolvedSecret` from the NDC SDK (which has also been removed from the SDK) with our own `Secret` type, which can be `Plain` or `FromEnvironment`. We then handle the case appropriately, whereas before, we would panic if we ever received a non-resolved secret.

I have introduced the `Environment` trait to supply environment variables. During normal execution, this is a wrapper around `sys::env::var`. In tests, however, we can use `EmptyEnvironment`, or a `HashMap` containing test data.

This also simplified a couple of pieces of test infrastructure.

* We no longer need to template the Aurora configuration (we just provide the appropriate value as an environment variable).
* Generating configuration becomes a single call to `ndc-postgres-cli update`.
* The benchmarks no longer need a generated configuration file.


justfile
```diff
@@ -165,28 +173,29 @@ test *args: start-dependencies create-aurora-ndc-metadata
...
+  CONNECTION_URI='{{POSTGRESQL_EMPTY_CONNECTION_URI}}' HASURA_PLUGIN_CONNECTOR_CONTEXT_PATH='{{POSTGRES_BROKEN_QUERIES_NDC_METADATA}}' \
+    cargo run --bin ndc-postgres-cli -- update
 
   # right now, we are breaking things, so archiving old configuration is meaningless
   # ./scripts/archive-old-ndc-metadata.sh '{{POSTGRES_V3_CHINOOK_NDC_METADATA}}'
 
-  ./scripts/generate-configuration.sh 'ndc-postgres' '{{POSTGRESQL_CONNECTION_STRING}}' '{{POSTGRES_V3_CHINOOK_NDC_METADATA}}/configuration.json'
-  ./scripts/generate-configuration.sh 'ndc-postgres' '{{CITUS_CONNECTION_STRING}}' '{{CITUS_V3_CHINOOK_NDC_METADATA}}/configuration.json'
-  ./scripts/generate-configuration.sh 'ndc-postgres' '{{COCKROACH_CONNECTION_STRING}}' '{{COCKROACH_V3_CHINOOK_NDC_METADATA}}/configuration.json'
+  CONNECTION_URI='{{POSTGRESQL_CONNECTION_URI}}' cargo run --bin ndc-postgres-cli -- --context='{{POSTGRES_V3_CHINOOK_NDC_METADATA}}' update
```
plcplc, line 182: It's quite nice how this bit of internal test infrastructure is now powered by a component that's part of the actual workflow of users!

plcplc (approved):
Looks good to me!

PR #334, opened 2024-03-01
CI: Publish releases automatically from tags.

SamirTalwar:
### What

When we tag a commit with a version number (e.g. `v1.2.34`), a new job will now:

1. Create a GitHub release (in draft).
2. Populate the release notes automatically from `changelog.md`.
3. Build CLI binaries with the version baked in.
4. Generate the connector metadata using the CLI.
5. Upload the metadata and CLI binaries as assets.

### How

Scripting shenanigans.

There's a new job to make the release, and a script that pulls out the relevant section from the changelog to build the release notes.

### DIFF
--- a/.github/workflows/ship.yaml
+++ b/.github/workflows/ship.yaml
@@ -171,6 +171,10 @@ jobs:
             declare "CARGO_TARGET_${TARGET_SCREAMING}_LINKER"='${{ matrix.linker }}'
             export "CARGO_TARGET_${TARGET_SCREAMING}_LINKER"
           fi
+          # If we're on a tag, use the tag as the release version.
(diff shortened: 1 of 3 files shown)


.github/workflows/ship.yaml
```diff
@@ -170,20 +171,57 @@ jobs:
             declare "CARGO_TARGET_${TARGET_SCREAMING}_LINKER"='${{ matrix.linker }}'
             export "CARGO_TARGET_${TARGET_SCREAMING}_LINKER"
           fi
+          # If we're on a tag, use the tag as the release version.
```
plcplc, line 174: Does this actually work? I would have thought you'd have to do something like `git tag --contains HEAD` or somesuch.

plcplc (approved):

PR #335, opened 2024-03-01
Update to spec-rc.19

paf31:
### What

Update to ndc-spec rc.19 in anticipation of tagging 0.1.0 today.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1472,6 +1472,23 @@ dependencies = [
 [[package]]
 name = "ndc-client"
 version = "0.1.0"
+source = "git+http://github.com/hasura/ndc-spec.git?tag=v0.1.0#8892f0524affd37e94097c2ce43da8740fc57aca"
+dependencies = [
+ "async-trait",
+ "indexmap 2.2.3",
+ "opentelemetry",
+ "reqwest",
+ "schemars",
+ "serde",
+ "serde_derive",
+ "serde_json",
+ "serde_with 2.3.3",
+ "url",
+]
+
+[[package]]
+name = "ndc-client"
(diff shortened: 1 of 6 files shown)


paf31:
The new benchmark output is kind of interesting too. There's quite a bit of variance in samples for each query, for some generated queries:

```
                         μ           Δ         σ       min       max
1a3e538c8464b282 2566.90μs ( +57.00μs) 2270.45μs 2070.00μs 3987.00μs 
1a8578f09c1ce296 1882.50μs (-101.70μs) 1890.44μs 1519.00μs 3070.00μs 
1a87a5bb4672f154 1653.50μs ( +75.70μs) 1592.56μs 1158.00μs 2726.00μs 
1a9a9e846616224e 11662.50μs (+122.50μs) 6217.37μs 10281.00μs 15604.00μs 
1ab272d91d113a24 2943.70μs (+124.20μs) 1094.24μs 2349.00μs 3527.00μs 
```

The standard deviation `σ` is comparable with the mean `μ` for many tests, and the `min-max` range is very wide. Running multiple benchmarks fails without a tolerance of at least 20% between runs on my machine.

SamirTalwar (dismissed):
Looks great, but I'm blocking this until the corresponding ndc-sdk PR is merged and the dependency is updated to something on `main`.

soupi:
I think this caught a query plan issue for cockroach where the roach did not support SELECT FROM clauses. This has since been [fixed](https://github.com/cockroachdb/cockroach/pull/116835). I'll see if pulling a newer version of cockroach fixes this issue.

soupi (approved):
thanks!

PR #341, opened 2024-03-04
Release: Publish SHA-256 checksums alongside CLI binaries.

SamirTalwar:
### What

In order to publish this CLI plugin in the index, we need SHA-256 checksums. We currently update the index manually, but let's generate the list while we're publishing so the poor soul updating the index doesn't have to download all the binaries to hash them.

### How

I have added a step to the release job that generates the checksums for all artifacts and stores them in the same directory, so they will be included.


soupi (approved):
very nice

PR #343, opened 2024-03-05
Read and write asynchronously from the file system.

SamirTalwar:
### What

Blocking a Tokio thread with `std::fs` is probably not wise in a server, so I have replaced it with `tokio::fs`, thanks to a kind suggestion from @hallettj. In the CLI and tests it's less of a big deal, but I have changed it regardless.

This has, conveniently, sped up a full test run from 15s to 10s on my MacBook. (On my Linux machine, the tests run very quickly even without this change.)

### How

[`tokio::fs`](https://docs.rs/tokio/latest/tokio/fs/index.html) is Tokio's asynchronous replacement for `std::fs`. It's mostly a drop-in replacement when using the utility functions such as `read_to_string`—just add `await`.

However, `fs::File::open` and `fs::File::create` now return an `AsyncRead` and `AsyncWrite` implementation respectively, which is not compatible with `serde_json::from_reader` and `serde_json::to_writer`. To work around this, I have replaced these operations with `fs::read_to_string` and `fs::write`, and converting to/from strings separately.

### DIFF
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -5,10 +5,10 @@
 
 mod metadata;
 
-use std::fs;
 use std::path::PathBuf;
 
 use clap::Subcommand;
+use tokio::fs;
 
 use ndc_postgres_configuration as configuration;
 use ndc_postgres_configuration::environment::Environment;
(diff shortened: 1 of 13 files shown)


crates/cli/src/lib.rs
```diff
@@ -118,11 +119,14 @@ async fn update(context: Context<impl Environment>) -> anyhow::Result<()> {
         .context_path
         .join(configuration::CONFIGURATION_FILENAME);
     let input: configuration::RawConfiguration = {
-        let reader = fs::File::open(&configuration_file_path)?;
-        serde_json::from_reader(reader)?
```
danieljharvey, line 122: Seems like `from_reader` is a bit slow and rubbish anyway, so this seems fine? https://github.com/serde-rs/json/issues/160

PR #349, opened 2024-03-07
Print errors at the top level using `Display`.

SamirTalwar:
### What

By default, errors are printed using their `Debug` implementation, which is unpleasant to read.

The errors have a `Display` implementation, which we can opt to use by capturing it and printing it explicitly.

### How

We capture the error, print it to STDERR with `eprintln!`, then return `ExitCode::FAILURE` (mapped to `EXIT_FAILURE` from libc, which is exit code 1 on POSIX systems).


soupi (approved):
thanks!

PR #351, opened 2024-03-11
Make `poolSettings` an explicit default

plcplc:
### What

We change the serialization policy of the `poolSettings` field such that the field is always serialized, regardless of its value.

This ensures that:

* It is possible for a user to pin the value, even the the pinned value is the default one.
* The defaults of the field are captured at project initialization time (and optionally when introspecting anew)

### DIFF
--- a/crates/configuration/src/version3/mod.rs
+++ b/crates/configuration/src/version3/mod.rs
@@ -30,7 +30,6 @@ pub const DEFAULT_CONNECTION_URI_VARIABLE: &str = "CONNECTION_URI";
 pub struct RawConfiguration {
     // Connection string for a Postgres-compatible database
     pub connection_uri: ConnectionUri,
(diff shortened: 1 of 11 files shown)


crates/configuration/src/version3/mod.rs
```diff
@@ -30,7 +30,6 @@ pub const DEFAULT_CONNECTION_URI_VARIABLE: &str = "CONNECTION_URI";
 pub struct RawConfiguration {
     // Connection string for a Postgres-compatible database
     pub connection_uri: ConnectionUri,
-    #[serde(skip_serializing_if = "PoolSettings::is_default")]
```
soupi, line 33: I assume is_default only exists for this particular setting, so if you'd like to remove it as well, you can.

soupi (approved):

PR #361, opened 2024-03-11
Rename schema fields and add a jsonschema generation and ref on init

soupi:
### What

This PR does the following:

1. It renames a few fields in the configuration:
   - `configureOptions` -> `introspectionOptions`
   - `mutationsVersion` moved from `configureOptions` to the top-level
   - connection uri, pool settings and isolation level are now nested under "connectionSettings"
   - Native queries use a new `ReadonlyColumnInfo` type instead of `ColumnInfo` type for arguments and column. ReadonlyColumnInfo does not contain default, identity and generated information which we don't want to appear in autocomplete for native queries.
2. It generates a `schema.jsonschema` file on cli `initialize` command next to the `configuration.json` file
3. It adds the `$schema` field to the configuration, referencing the generated `schema.jsonschema` file.

The purpose of (2) and (3) is to provide auto-complete features when editing the configuration in vscode (demo soon).

Screenshot from 2024-03-12 16-56-07


justfile
```diff
@@ -79,10 +79,6 @@ dev-citus: start-dependencies
     -x clippy \
     -x 'run --bin ndc-postgres -- serve --configuration {{CITUS_V3_CHINOOK_NDC_METADATA}}'
 
-# Generate the JSON Schema document for the the configuration.
-document-jsonschema:
```
plcplc, line 83: This no longer exists?

static/aurora/v3-chinook-ndc-metadata/configuration.json
```diff
@@ -2813,7 +2816,7 @@
         "operatorKind": "custom"
       }
     ],
-    "mutationsVersion": null,
+    "mutationsVersion": "v1",
```
plcplc, line 2819: Should mutationsVersion change to "v1"?

crates/configuration/src/configuration.rs
```diff
@@ -79,9 +80,9 @@ pub async fn parse_configuration(
         }?;
     Ok(Configuration {
         metadata: configuration.metadata,
-        pool_settings: configuration.pool_settings,
+        pool_settings: configuration.connection_settings.pool_settings,
         connection_uri,
-        isolation_level: configuration.isolation_level,
-        mutations_version: configuration.configure_options.mutations_version,
+        isolation_level: configuration.connection_settings.isolation_level,
+        mutations_version: configuration.mutations_version,
```
SamirTalwar, line 86: Perhaps it's worth just storing `connection_settings` directly in the `Configuration`.

SamirTalwar (commented):
I like this a lot.

plcplc (approved):
Looks good to me.

With this PR the configuration fields are better grouped according to their usage.

IMO only `metadata.nativeQueries` stands out a bit, since it's authored by users while the rest of the `metadata` object is generated. Whether we should do anything about that or not is a separate issue.

static/schema.jsonschema
plcplc, line 1: What generates this file?

crates/configuration/src/configuration.rs
soupi: Thought about it, but didn't want to make too many changes, and wasn't sure about whether we should use version3 structure here. Maybe we can decide in a separate PR since this is internal to the connector?

static/schema.jsonschema
soupi: It was a manual copy paste tbh. Maybe we can decide later how to handle this for local dev?

plcplc: If the only purpose is to have something for the $schema field to reference I would be fine with just eliding the file

plcplc: That is, the schema field in the test-configurations

soupi: Yes that's exactly it's use.

crates/configuration/src/configuration.rs
soupi: In general I do agree that that's a better structure though.

PR #362, opened 2024-03-14
update command should not override the config file in some cases

soupi:
### What

We want the update command in ndc-postgres-cli to not override changed made after we started introspecting but before we wrote the changes to file, and not to write to the filesystem when the file does not need changing.

### How

1. We loop the update operation 3 times
2. We read the input file after the introspection took place as well
3. If the input from before the introspection and the input from after the introspection are the same we continue, otherwise we try again
4. Before writing to file, we check that if the output is going to be the same as the input, if it is the same, we skip writing, if not, we write.
5. We add the `Eq` trait everywhere so we can compare the config.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -7,6 +7,12 @@
 - The ndc-postgres-cli `initialize` command will now generate
   the jsonschema of the configuration format as well.
   ([#361](https://github.com/hasura/ndc-postgres/pull/361))
+- The ndc-postgres-cli `update` command will check that the configuration file has
+  not changed in the middle of the introspection process, and will retry if it did.
+  ([#362](https://github.com/hasura/ndc-postgres/pull/362))
+- The ndc-postgres-cli `update` command will not write the introspection result to
+  file if the configuration file is already up to date.
+  ([#362](https://github.com/hasura/ndc-postgres/pull/362))
 - A few fields in the configuration format has been changed:
 
   - `configureOptions` was renamed to `introspectionOptions`
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -13,6 +13,8 @@ use tokio::fs;
 use ndc_postgres_configuration as configuration;
 use ndc_postgres_configuration::environment::Environment;
 
+const UPDATE_ATTEMPTS: u8 = 3;
+
 /// The various contextual bits and bobs we need to run.
 pub struct Context<Env: Environment> {
     pub context_path: PathBuf,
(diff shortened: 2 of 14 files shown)


crates/cli/tests/initialize_tests.rs
soupi, line 1: a few tests for the previous change, adding the schema.json.

crates/configuration/src/configuration.rs
```diff
@@ -17,7 +17,7 @@ pub const CONFIGURATION_FILENAME: &str = "configuration.json";
 pub const CONFIGURATION_JSONSCHEMA_FILENAME: &str = "schema.json";
 
 /// The parsed connector configuration.
-#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema)]
+#[derive(Clone, PartialEq, Eq, Debug, Deserialize, Serialize, JsonSchema)]
```
SamirTalwar, line 20: The lack of a consistent order here is going to bug me for the rest of my days.

SamirTalwar (commented):
This seems wise.

I'll review again when it's ready.

crates/configuration/src/configuration.rs
soupi: This should probably be an auto formatting option :)

changelog.md
```diff
@@ -7,6 +7,12 @@
 - The ndc-postgres-cli `initialize` command will now generate
   the jsonschema of the configuration format as well.
   ([#361](https://github.com/hasura/ndc-postgres/pull/361))
+- The ndc-postgres-cli `update` command will check that the configuration file has
+  not changed in the middle of the introspection process, and will retry if it did.
+  ([#362](https://github.com/hasura/ndc-postgres/pull/362))
+- The ndc-postgres-cli `update` command will not write the introspection result to
+  file if the configuration file is already up to date.
+  ([#362](https://github.com/hasura/ndc-postgres/pull/362))
```
plcplc, line 15: Consider merging these entries. It's a bit weird to have two changelog entries for one PR.

crates/cli/src/lib.rs
```diff
@@ -128,30 +130,63 @@ async fn initialize(with_metadata: bool, context: Context<impl Environment>) ->
...
-                    )
-                } else {
-                    anyhow::anyhow!(err)
-                }
-            })?;
-        serde_json::from_str(&configuration_file_contents)?
-    };
-    let output = configuration::introspect(input, &context.environment).await?;
+    // It is possible to change the file in the middle of introspection.
+    // We want to detect these scenario and try again, or fail if we are unable to.
```
plcplc, line 134:
```suggestion
    // We want to detect this scenario and retry, or fail if we are unable to.
```

crates/cli/src/lib.rs
```diff
@@ -128,30 +130,63 @@ async fn initialize(with_metadata: bool, context: Context<impl Environment>) ->
...
+            if input != output {
+                fs::write(
+                    &configuration_file_path,
+                    serde_json::to_string_pretty(&output)?,
+                )
+                .await?;
+            } else {
+                println!("The configuration is up-to-date. Nothing to do.");
+            }
+            return Ok(());
```
plcplc, line 167: Does this actually exit the loop?

plcplc (approved):

crates/cli/src/lib.rs
SamirTalwar: If the tests pass, yes. 😁

PR #366, opened 2024-03-15
fix names of fields in schema generation

soupi:
### What

The `/schema` endpoint was exposing the wrong column names in object types. This PR fixes this.

### How

columns, referenced by tables, native queries, etc. Have a type that looks somewhat like this: `BTreeMap<String, ColumnInfo>`.

`ColumnInfo` looks somewhat like this: `ColumnInfo { name: String, ... }`.

The `String` in the map refers to the external name that ndc-spec queries should reference, while the `ColumnInfo` name refers to the database column name.

By mistake, were reporting the `ColumnInfo` name in the schema, instead of the String from the map.

We fix this by referring to the right name, and use `iter()` on the BTreeMap instead of `values()`.

### DIFF
--- a/crates/connectors/ndc-postgres/src/schema.rs
+++ b/crates/connectors/ndc-postgres/src/schema.rs
@@ -186,52 +186,59 @@ pub async fn get_schema(
         BTreeMap::from_iter(metadata.tables.0.iter().map(|(collection_name, table)| {
             let object_type = models::ObjectType {
                 description: table.description.clone(),
-                fields: BTreeMap::from_iter(table.columns.values().map(|column| {
-                    (
-                        column.name.clone(),
-                        models::ObjectField {
-                            description: column.description.clone(),
-                            r#type: column_to_type(column),
-                        },
-                    )
-                })),
+                fields: BTreeMap::from_iter(table.columns.iter().map(
+                    |(column_name, column_info)| {
+                        (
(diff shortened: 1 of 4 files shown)


SamirTalwar:
I wanted to wrap these all in newtypes to catch exactly this kind of issue (we saw something similar before) but I struggled because a lot of this stuff is in the `ndc_client::models` module.

I still think we should do this, but it requires a bit of thought.

crates/connectors/ndc-postgres/src/schema.rs
```diff
@@ -186,52 +186,59 @@ pub async fn get_schema(
         BTreeMap::from_iter(metadata.tables.0.iter().map(|(collection_name, table)| {
```
SamirTalwar, line 186: FWIW, `BTreeMap::from_iter` might be more idiomatically expressed as `.collect()`.

SamirTalwar (approved):

crates/connectors/ndc-postgres/src/schema.rs
soupi: good to know!

soupi:
@SamirTalwar Perhaps we should rename the field `name` in column info to `pg_name`.

soupi:
@SamirTalwar ah, that's a breaking change, and I don't care to make one right now :(

PR #372, opened 2024-03-20
Read Native Query SQL from files

soupi:
### What

Currently we can specify native queries by inlining their SQL as a string in the `sql` field.
We'd like to be able to reference a `.sql` file instead, so we can avoid escaping characters and can use newlines.

The new config format will accept the following structures:

```json
"sql": "<inline sql>" (as before, but also:)

"sql": { "inline": "<inline sql>" }

"sql": { "file": "<relative file path>" }
```

### How

- We parse the `NativeQuerySql` which represents the SQL field using an intermediate new type `NativeQuerySqlExternal`.
- as part of the parse_configuration step and after serde deserialization, we traverse the metadata and convert `NativeQuerySqlExternal` to `NativeQuerySql` by reading from file (performing IO) if needed, and parsing to `NativeQueryParts`. We use the configuration directory that the user supplied to locate the relative path.

We replace the existing `sql` field in nativequeryinfo with an Either external or internal representation. This is not ideal because it breaks the parse don't validate principle.
We would like to mitigate this in the future by separating the `RawConfiguration` types and the (execution time) `Configuration` types completely.
We will do that as part of a larger ticket of redesigning the configuration format so it can be split across multiple files.


crates/configuration/src/configuration.rs
```diff
@@ -60,14 +60,29 @@ pub async fn parse_configuration(
...
 ) -> Result<Configuration, Error> {
     let configuration_file = configuration_dir.as_ref().join(CONFIGURATION_FILENAME);
-    let configuration_file_contents = fs::read_to_string(&configuration_file).await?;
-    let configuration: version3::RawConfiguration =
+
+    let configuration_file_contents =
+        fs::read_to_string(&configuration_file)
+            .await
+            .map_err(|err| {
+                Error::IoErrorButStringified(format!("{}: {}", &configuration_file.display(), err))
```
soupi, line 68: I just want rust to tell me which file it couldn't read, so I added these everywhere. Sorry.

crates/configuration/src/configuration.rs
```diff
@@ -60,14 +60,29 @@ pub async fn parse_configuration(
...
+            })?;
+    let mut configuration: version3::RawConfiguration =
         serde_json::from_str(&configuration_file_contents).map_err(|error| Error::ParseError {
             file_path: configuration_file.clone(),
             line: error.line(),
             column: error.column(),
             message: error.to_string(),
         })?;
+    // look for native query sql file references and read from disk.
+    for native_query_sql in configuration.metadata.native_queries.0.values_mut() {
```
soupi, line 78: step after serialization - read from disk.

crates/query-engine/metadata/src/metadata/native_queries.rs
```diff
@@ -75,42 +214,13 @@ impl Serialize for NativeQuerySql {
                 }
             }
         }
-        serializer.serialize_str(&sql)
-    }
-}
-
-struct NQVisitor;
```
soupi, line 82: @plcplc taught me that we can avoid all of this cruft by just implementing From/Into.

crates/query-engine/metadata/src/metadata/native_queries.rs
```diff
@@ -139,57 +258,79 @@ fn parse_native_query(string: &str) -> Vec<NativeQueryPart> {
...
                 NativeQueryPart::Text(" = '{name}'".to_string()),
-            ]
+            ])
+        );
+    }
+
+    #[test]
+    fn parse_inline_untagged() {
+        assert_eq!(
+            serde_json::from_str::<NativeQuerySqlExternal>(r#""select 1""#).unwrap(),
```
soupi, line 320: We still want to be able to parse untagged inline expressions just as before.

crates/query-engine/translation/src/translation/query/native_queries.rs
```diff
@@ -31,6 +31,8 @@ pub fn translate(env: &Env, state: State) -> Result<Vec<sql::ast::CommonTableExp
         let sql: Vec<sql::ast::RawSql> = native_query
             .info
             .sql
+            .sql()
+            .map_err(Error::InternalError)?
```
soupi, line 35: This might fail because of our types shape, we'll fix this in the future.

crates/tests/databases-tests/src/postgres/snapshots/databases_tests__postgres__configuration_tests__get_rawconfiguration_v3_schema.snap
```diff
@@ -819,7 +819,45 @@ expression: schema
         }
       }
     },
-    "Native_query_sql": {
+    "NativeQuerySql": {
```
soupi, line 822: The new schema for the sql field.

crates/tests/tests-common/src/ndc_metadata/configuration.rs
```diff
@@ -15,30 +15,68 @@ use super::helpers::get_path_from_project_root;
...
+    let temp_deploys_path = get_path_from_project_root(temp_deploys_path);
+
+    // make sure the directory where all temp deploys are copied to exists
+    fs::create_dir_all(&temp_deploys_path)
+        .await
+        .map_err(|err| anyhow::anyhow!("{}: {}", &temp_deploys_path.display(), err))?;
+
+    let new_ndc_metadata_dir = temp_deploys_path.join(db_name);
+
+    copy_dir::copy_dir(&ndc_metadata_dir_path, &new_ndc_metadata_dir).map_err(|err| {
```
soupi, line 41: We now also copy the entire configuration directory.

static/postgres/v3-chinook-ndc-metadata/configuration.json
```diff
@@ -1484,7 +1502,7 @@
         "description": "A native query used to test support for composite types"
       },
       "summarize_organizations": {
-        "sql": "SELECT 'The organization ' || org.name || ' has ' || no_committees::text || ' committees, ' || 'the largest of which has ' || max_members || ' members.' AS result FROM (SELECT orgs.* FROM unnest({{organizations}}) as orgs) AS org JOIN LATERAL ( SELECT count(committee.*) AS no_committees, max(members_agg.no_members) AS max_members FROM unnest(org.committees) AS committee JOIN LATERAL ( SELECT count(*) as no_members FROM unnest(committee.members) AS members) AS members_agg ON true) AS coms ON true",
+        "sql": { "file": "./native_queries/summarize_organizations.sql" },
```
soupi, line 1505: New thing!

static/postgres/v3-chinook-ndc-metadata/native_queries/summarize_organizations.sql
soupi, line 1: New thing!

crates/query-engine/metadata/src/metadata/native_queries.rs
```diff
@@ -46,6 +47,147 @@ pub struct ReadOnlyColumnInfo {
...
+    /// used for deserialization.
+    fn from(value: NativeQuerySql) -> Self {
+        match value {
+            NativeQuerySql::Inline { sql } => NativeQuerySqlExternal::Inline { inline: sql },
+            NativeQuerySql::FromFile { file, .. } => NativeQuerySqlExternal::File { file },
+        }
+    }
+}
+
+impl JsonSchema for NativeQuerySqlEither {
```
plcplc, line 181: If you want to, you can avoid this schema impl with the [`#[schemars(with="NativeQuerySqlExternal")]`](https://graham.cool/schemars/deriving/attributes/#with) attribute on `NativeQuerySqlEither` and [`#[schemars(rename="NativeQuerySql")]`](https://graham.cool/schemars/deriving/attributes/#rename) on `NativeQuerySqlExternal`.

Although it's arguably mostly a matter of style for this example.

crates/query-engine/metadata/src/metadata/native_queries.rs
plcplc: 🌺

crates/query-engine/metadata/src/metadata/native_queries.rs
```diff
@@ -55,62 +197,38 @@ pub enum NativeQueryPart {
...
-    fn deserialize<D>(deserializer: D) -> Result<NativeQuerySql, D::Error>
-    where
-        D: serde::Deserializer<'de>,
-    {
-        deserializer.deserialize_str(NQVisitor)
-    }
-}
-
-impl JsonSchema for NativeQuerySql {
+impl JsonSchema for NativeQueryParts {
```
plcplc, line 229: You can probably also avoid this instance with a `#[schemars(rename=..)]` attribute.

static/postgres/v3-chinook-ndc-metadata/native_queries/summarize_organizations.sql
```diff
@@ -0,0 +1,22 @@
+SELECT
```
plcplc, line 1: Ah, how I wished I could have had a separate file when I made this query. Boy is it easier on the eyes without string escaping~

plcplc (approved):
Good stuff! A few easy (stylistic) improvements in comments.

We'll get back to parse-don't-validate eventually 🌺

crates/query-engine/metadata/src/metadata/native_queries.rs
```diff
@@ -46,6 +47,147 @@ pub struct ReadOnlyColumnInfo {
...
+            },
+        }
+    }
+}
+
+impl From<NativeQuerySql> for NativeQuerySqlExternal {
+    /// used for deserialization.
+    fn from(value: NativeQuerySql) -> Self {
+        match value {
+            NativeQuerySql::Inline { sql } => NativeQuerySqlExternal::Inline { inline: sql },
```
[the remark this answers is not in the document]
soupi: Yes, this is intended so that we could phase out raw string eventually.

crates/query-engine/metadata/src/metadata/native_queries.rs
soupi: I think this doesn't work because it attempts to create a schema for `Vec<NativeQueryPart>`, even if I add `with = "String"`.

crates/query-engine/metadata/src/metadata/native_queries.rs
```diff
@@ -120,9 +240,18 @@ impl JsonSchema for NativeQuerySql {
...
 // Parsing
 
-/// Parse a native query into parts where variables have the
-/// syntax `{{<variable>}}`.
-fn parse_native_query(string: &str) -> Vec<NativeQueryPart> {
+/// Read a file a parse it into native query parts.
+pub fn parse_native_query_from_file(file: std::path::PathBuf) -> Result<NativeQuerySql, String> {
+    let contents: String = match fs::read_to_string(&file) {
+        Ok(ok) => Ok(ok),
+        Err(err) => Err(format!("{}: {}", &file.display(), err)),
```
SamirTalwar, line 247: Again, here seems like a good place for `Error::IoError { path: Option<PathBuf>, inner: io::Error }`.

SamirTalwar (commented):
Sorry for the delay here. I only have some minor comments on delaying error stringification.

crates/query-engine/metadata/src/metadata/native_queries.rs
soupi: I would approve such a PR :)

PR #374, opened 2024-03-21
When writing configuration files, always end them in a newline.

SamirTalwar:
### What

This is a personal pet peeve, mostly brought on by Git highlighting this in diffs.

### How

I have scattered `+ "\n"` throughout the code, and added some test assertions to match.

### DIFF
--- a/crates/cli/tests/common/mod.rs
+++ b/crates/cli/tests/common/mod.rs
@@ -1,5 +1,5 @@
 #![allow(dead_code)] // required because this is included mulitple times
-                     //
+
 use std::path::Path;
 
 use tokio::fs;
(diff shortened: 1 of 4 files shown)


crates/cli/tests/common/mod.rs
```diff
@@ -0,0 +1,15 @@
+#![allow(dead_code)] // required because this is included mulitple times
```
soupi, line 1: I don't understand?

soupi (approved):
Ok.

Btw and unrelated to this, I think we might want to have our own "to_json_string"/parsing methods instead of using serde_json directly, so we can make changes like this across the codebase more easily.

crates/cli/tests/common/mod.rs
soupi: weird.

PR #378, opened 2024-03-22
Use Rust 1.77.0

danieljharvey:
### What

Upgrade to use Rust 1.77.0

https://github.com/rust-lang/rust/releases/tag/1.77.0

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -116,7 +116,6 @@
               pkgs.cargo-machete
               pkgs.cargo-nextest
               pkgs.cargo-watch
-              pkgs.rnix-lsp
               rust.rustToolchain
 
               # Benchmarks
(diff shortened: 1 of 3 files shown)


flake.nix
```diff
@@ -116,7 +116,6 @@
               pkgs.cargo-machete
               pkgs.cargo-nextest
               pkgs.cargo-watch
-              pkgs.rnix-lsp
```
paritosh-08, line 119: 👀

paritosh-08 (approved):

danieljharvey:
`cargo machete` is having a bad time now, maybe a bug? https://github.com/hasura/ndc-postgres/actions/runs/8393753764/job/22989435799?pr=378#step:5:12

flake.nix
danieljharvey: It's been marked as deprecated in `nixpkgs`, so removing for now.

PR #379, opened 2024-03-22
NativeQuery refac leftovers

plcplc:
### What

Tidy up a few leftovers from the recent change to have native queries able to live in files.

### How

`just generate-configuration`


danieljharvey (approved):
Aha!

PR #380, opened 2024-03-22
Add support for introspecting domain types

plcplc:
### What

This PR adds support for introspecting domain types.

### How

Prior to this PR domain types were already recognised as scalar types. However, the ability to view a domain type transparently as its base type was not, so you couldn't _do_ anything with values of domain type.

This PR adds cases to the set of implicit casts for domain types in the introspection query, which enables support for comparison operators.

### DIFF
--- a/static/schema.json
+++ b/static/schema.json
@@ -716,7 +716,7 @@
           "description": "SQL expression to use for the Native Query. We can interpolate values using `{{variable_name}}` syntax, such as `SELECT * FROM authors WHERE name = {{author_name}}`",
           "allOf": [
             {
-              "$ref": "#/definitions/Native_query_sql"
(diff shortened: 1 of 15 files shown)


static/schema.json
```diff
@@ -745,7 +745,41 @@
         }
       }
     },
-    "Native_query_sql": {
+    "NativeQuerySql": {
```
danieljharvey, line 748: This feels like it's related to @soupi 's recent `.sql` files change for native queries rather than domain types?

danieljharvey (approved):
Lovely that all we have to really do is say "X is actually Y" and then everything else falls into place.

SamirTalwar (commented):
Late review, but I just wanted to say, thank you for teaching me something. 😊

PR #381, opened 2024-03-22
Cast-extend aggregation functions

plcplc:
### What

This PR makes the introspection of aggregation functions take implicit casts into account.

This enables aggregation functions on domain types, which would typically only be defined on the base type.

It also enables certain other aggregations for scalar types that weren't defined before, most notably aggregation functions for `float4` on CockroachDB, which becomes accessible through implicit casts to `float8`.

### How

The cast-extension for aggregation functions introspection code is based on the same for comparison operators.

We also change our notion of "occurring scalar types" to include the return types aggregation functions. Previously, aggregation functions could be filtered out if their return type didn't feature in the set of occurring scalar types, which was unnecessarily strict.


crates/configuration/src/version3/version3.sql
```diff
@@ -360,6 +440,65 @@ WITH
...
+        -- Therefore we resort to filtering by ordered ROW_NUMBER().
+        ROW_NUMBER()
+          OVER
+          (
+            PARTITION BY
+              proc_name, argument_type
+            ORDER BY
+              -- Prefer uncast argument.
+              NOT argument_casted DESC,
+              -- Arbitrary desperation: Lexical ordering
```
danieljharvey, line 486: 🤣

danieljharvey (approved):
Absolutely great.

PR #387, opened 2024-03-25
Support enum types

plcplc:
### What

This PR improves introspection support for enum types.

Comparison operators and aggregation functions on enum types are implicitly defined in PostgreSQL and thus were not detected prior to this PR.

### How

We simply collect which enum types are defined and make up the conventional comparison operators and the min/max aggregation functions for all enum types.

We test that we can select and aggregate enum types, and that we can insert enum values which PostgreSQL validates.

We also collect the enum value labels, which we will be using once ndc-spec v0.1.1 is released.

### DIFF
--- a/crates/tests/databases-tests/src/postgres/snapshots/databases_tests__postgres__mutation_tests__negative__v1_insert_enum_invalid_label.snap
+++ b/crates/tests/databases-tests/src/postgres/snapshots/databases_tests__postgres__mutation_tests__negative__v1_insert_enum_invalid_label.snap
@@ -1,0 +1,10 @@
+---
+source: crates/tests/databases-tests/src/postgres/mutation_tests.rs
+expression: result
+---
+{
+  "message": "Unprocessable content",
+  "details": {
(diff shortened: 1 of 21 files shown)


crates/tests/databases-tests/src/postgres/snapshots/databases_tests__postgres__mutation_tests__negative__v1_insert_enum_invalid_label.snap
```diff
@@ -0,0 +1,10 @@
+---
+source: crates/tests/databases-tests/src/postgres/mutation_tests.rs
+expression: result
+---
+{
+  "message": "Unprocessable content",
+  "details": {
+    "detail": "error returned from database: invalid input value for enum card_suit: \"horses\""
```
danieljharvey, line 8: 🐴

plcplc: 🐣

danieljharvey (approved):
Absolutely great.

PR #390, opened 2024-03-26
Upgrade ndc-spec to v0.1.1.

SamirTalwar:
### What

This includes:

- type representations in the models
- upgrading the OpenTelemetry SDK
- adding a feature switch to support `rustls` instead of `native-tls`
- a lot of enhancements to ndc-test

ndc-sdk was upgraded in https://github.com/hasura/ndc-sdk-rs/pull/1.

We do not actually specify type representations in this changeset. We will add these in a follow-up change.

### How

I discovered we can specify workspace dependencies, which means the versions only need to go in one place. I am experimenting with this Cargo feature in this changeset. If we like it, we may want to use it for all dependencies.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,3 +16,8 @@ members = [
 
 [workspace.lints.clippy]
 all = "warn"
+
+[workspace.dependencies]
+ndc-client = { git = "https://github.com/hasura/ndc-spec.git", tag = "v0.1.1" }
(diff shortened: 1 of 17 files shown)


Cargo.toml
```diff
@@ -16,3 +16,8 @@ members = [
 
 [workspace.lints.clippy]
 all = "warn"
+
+[workspace.dependencies]
+ndc-client = { git = "https://github.com/hasura/ndc-spec.git", tag = "v0.1.1" }
```
danieljharvey, line 21: 👀

PR #391, opened 2024-03-27
Support introspecting composite types

plcplc:
### What

This PR adds the ability of the introspection query to detect composite types, meaning user-defined record types that do not arise from a table but exist as standalone definitions.

This even discovered a typo in the previous, manually crafted composite type metadata.

Support for cockroachdb is limited here until https://github.com/cockroachdb/cockroach/issues/109675 is released.

### How

**Introspection Query** now captures composite types, filtering out the tables. Making metadata json out of these is quite similar to how we already did tables, so no huge surprises here.

**Occurring types logic** now becomes somewhat more complex because we can no longer simply look at which type names are used in the collections we track. A composite type occurring in say, a table column, may refer to scalar types that don't occur anywhere else, and even other composite types. Occurring type discovery thus becomes an iterative procedure rather than a single pass.

### DIFF
--- a/crates/configuration/src/version3/mod.rs
+++ b/crates/configuration/src/version3/mod.rs
@@ -162,14 +162,14 @@ pub async fn introspect(
 
     // We filter our comparison operators and aggregate functions to only include those relevant to
     // types that may actually occur in the schema.
-    //
-    // We don't (yet) filter composite types since that would require iterating over the set of
-    // relevant types until reaching a fix point. (The same would strictly be true for aggregation
-    // functions, if we supported composing them)
     let relevant_comparison_operators =
         filter_comparison_operators(&scalar_types, comparison_operators);
     let relevant_aggregate_functions =
         filter_aggregate_functions(&scalar_types, aggregate_functions);
+    let relevant_composite_types = transitively_occurring_composite_types(
+        occurring_composite_types(&tables, &args.metadata.native_queries),
+        composite_types,
+    );
 
     Ok(RawConfiguration {
         schema: args.schema,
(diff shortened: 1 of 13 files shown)


crates/configuration/src/version3/mod.rs
```diff
@@ -170,14 +179,103 @@ pub async fn introspect(
...
+    }
+
+    // Since 'discovered_type_names' only grows monotonically starting from 'occurring_type_names'
+    // we just have to compare the number of elements to know if new types have been discovered.
+    if discovered_type_names.len() == occurring_type_names.len() {
+        // Iterating over occurring types discovered no new types
+        composite_types
+            .0
+            .retain(|t, _| occurring_type_names.contains(t));
+        (occurring_scalar_types, composite_types)
```
SamirTalwar, line 266: Why do we both mutate `composite_types` and return it?

crates/configuration/src/version3/mod.rs
```diff
@@ -170,14 +179,103 @@ pub async fn introspect(
...
+        .collect::<BTreeSet<String>>()
+}
+
+// Since array types and composite types may refer to other types we have to transitively discover
+// the full set of types that are relevant to the schema.
+pub fn transitively_occurring_types(
+    mut occurring_scalar_types: BTreeSet<metadata::ScalarType>,
+    occurring_type_names: BTreeSet<String>,
+    mut composite_types: metadata::CompositeTypes,
+) -> (BTreeSet<metadata::ScalarType>, metadata::CompositeTypes) {
```
SamirTalwar, line 227: The mix of mutated "in/out" parameters and return values is confusing me. I'd rather pick one or the other.

Can we just mutate `occurring_type_names`?

crates/configuration/src/version3/version3.sql
SamirTalwar, line 1: I cannot promise I completely understand this SQL, but I understand it about as well as I'm ever going to.

SamirTalwar (approved):

crates/configuration/src/version3/version3.sql
plcplc: That's understandable. It's mostly made up of semi-internal postgres details :p

PR #392, opened 2024-03-27
Iteratively apply implicit casts

plcplc:
### What

This PR improves the handling of implicit casts during introspection by computing the transitive closure of implicit casts.

This is necessary to e.g. support any domain types defined over `varchar`.
From `domain-types.sql`:

```
-- This type tests that we consider implicit casts iteratively. If we didn't
-- "Phone" wouldn't have any comparison operators defined. with iterative
-- implicit casting, Phone gets to inherit the oprators of "text" via the
-- successive casts "Phone -> varchar -> text".
CREATE DOMAIN "Phone" varchar(20);
```

This type is used in the AdventureWorks sample database. Without this change we don't generate any comparison operators for `Phone`, which means e.g. it cannot participate in relationships.

### How

In order to compute the transitive closure we use a recursive CTE statement, where we combinatorically collect cast-arrows, taking care to stop whenever we encounter a cycle, which may occur (e.g., `varchar` and `text` values are completely interchangeable).

In addition we record the path of types each transitive cast takes, which we can now use for debugging as well as preferring "less casted" variants over "more casted" variants (of comparison functions for example) rather than simply "casted or not" variants.

In making this I discovered that both cast paths `varchar -> text` and `varchar -> bpchar` exist, and many of the same functions are defined for both `text` and `bpchar`. Because we use lexicographic sorting as a tie-breaker we would sometimes wind up functions taking `bpchar` arguments. Since `text` seems like a much more conventional choice most times I simply elected to censor the `varchar -> bpchar` and `text -> bpchar` casts.

Another side-effect of this is that, because a few functions on geography/geometry types such as `st_covers` etc. are also defined for `text` _directly_ they now become available on all types that are iteratively implicitly castable to text. This is an issue to address later however.

### DIFF
--- a/.envrc.example
+++ b/.envrc.example
@@ -13,4 +13,4 @@ export CARGO_INCREMENTAL='true'
 export RUSTFLAGS='-C link-arg=-fuse-ld=mold'
 
 # set your AWS Aurora connection string for testing
-export AURORA_CONNECTION_STRING='postgresql://...'
+export AURORA_CONNECTION_URI='postgresql://...'
--- a/.github/workflows/cargo-test.yaml
+++ b/.github/workflows/cargo-test.yaml
@@ -140,7 +140,7 @@ jobs:
           cargo nextest run --no-fail-fast --release --features aurora --filter-expr='package(databases-tests)'
         env:
           RUST_LOG: INFO
-          AURORA_CONNECTION_STRING: ${{ secrets.AURORA_CONNECTION_STRING }}
+          AURORA_CONNECTION_URI: ${{ secrets.AURORA_CONNECTION_URI }}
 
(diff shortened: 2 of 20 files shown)


soupi (approved):
Can't say I completely understand the new sql, but it makes sense, the generated config looks good, and I can verify that we can introspect the Phone type from the adventureworks db.

soupi:
I will release a new cli version when I get back.

plcplc:
Yeah, the SQL uses the `WITH RECURSIVE` CTE form to iteratively apply the same query on the set of implicit casts. The execution model for that is a bit unusual, and it took a bit of fiddling to get right.

PR #393, opened 2024-03-28
experimental mutations version with boolean expressions

soupi:
### What

We introduce a new experimental mutations version. This version is intended to be used as a WIP version not for public consumption.
It also introduces an update delete mutation that contains a `filter` predicate field.

### How

1. Introduce a new mutations version: `VeryExperimentalWip`.
2. Take the existing `insert`, `delete`, `translate` and `generate` modules, duplicate them, and place them in `v1/` and `experimental/` directories.
3. Add and modify existing code to dispatch on operations from the right module according to the mutations version in the config.
4. Add a new `filter` field for the delete operation which will be interpreted as a boolean expression and added as a `WHERE` clause in the delete sql.
5. Expose this new field as part of the schema as a boolean expression over the collection.


crates/tests/tests-common/goldenfiles/mutations/delete_invoice_line.json
```diff
@@ -4,7 +4,25 @@
...
+                "arguments": {}
+              }
+            ]
+          },
+          "operator": "_eq",
+          "value": {
+            "type": "scalar",
+            "value": 512
+          }
+        }
```
soupi, line 25: this is new.

crates/tests/databases-tests/src/postgres/snapshots/databases_tests__postgres__explain_tests__mutation__delete_invoice_line.snap
```diff
@@ -0,0 +1,51 @@
...
+      FROM
+        "public"."Track" AS "%3_BOOLEXP_Track"
+      WHERE
+        (
+          (
+            "%2_InvoiceLine"."TrackId" = "%3_BOOLEXP_Track"."TrackId"
+          )
+          AND (
+            ("%2_InvoiceLine"."InvoiceLineId" = 90)
+            AND ("%3_BOOLEXP_Track"."TrackId" = 512)
```
soupi, line 21: this is new

crates/query-engine/metadata/src/metadata/mutations.rs
```diff
@@ -8,4 +8,5 @@ use serde::{Deserialize, Serialize};
 #[serde(rename_all = "camelCase")]
 pub enum MutationsVersion {
     V1,
+    VeryExperimentalWip,
```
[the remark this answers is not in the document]
plcplc: It's excellent that we can iterate on the design in this way.

Maybe it would have been better to pick a different name for these new experimental versions though. Consider which name we should pick the next time we want to try a different experiment. `VeryExperimentalWip` is a bit reminiscient of document file name versions like `report_final.doc` and report_final_2.doc` if you know the meme :-).

crates/query-engine/translation/src/translation/mutation/generate.rs
```diff
@@ -1,35 +1,33 @@
...
 
+use super::experimental;
+use super::v1;
+
 #[derive(Debug, Clone)]
 pub enum Mutation {
-    DeleteMutation(DeleteMutation),
-    InsertMutation(InsertMutation),
+    V1(v1::Mutation),
+    Experimental(experimental::Mutation),
```
plcplc, line 12: Shouldn't this also be called `VeryExperimentalWip`?

plcplc (approved):
Good stuff. The PR comments were a great aid in reviewing!

crates/query-engine/translation/src/translation/mutation/generate.rs
soupi: The name of the flag is deliberately exaggerated, but we don't have to do that internally :)

I assume we won't want to have two distinct versions at the same time, but if we do, we can give them different names, sure.
This `VeryExperimentalWip` version should be renamed to `V2` once we fix it after all.

crates/query-engine/metadata/src/metadata/mutations.rs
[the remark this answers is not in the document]
danieljharvey: In future, perhaps something like `unsafe_next`? We only ever want one of these versions at once, and it communicates that if you use it, all stability bets are off?

PR #395, opened 2024-03-28
bring back archived configs

soupi:
### What

We no longer break compatibility, so it's time to bring back the test that checks that we can still read older configurations.

### How

The script now has to keep a directory rather than a file.


static/aurora/v3-chinook-ndc-metadata/configuration.json
soupi, line 1: this wasn't up to date.

SamirTalwar:
Does this test make sense now that the configuration can be multiple files?

soupi:
@SamirTalwar yes. It only checks that the configuration format can be parsed by the connector, and the configuration.json is still a single file.

SamirTalwar (approved):

PR #397, opened 2024-03-29
Add enum type representations

soupi:
### What

ndc-spec 0.1.1 introduce an [optional type representation hint](https://hasura.github.io/ndc-spec/specification/schema/scalar-types.html#type-representations). In this PR we provide type representation for enum values.

### How

1. We add a new `TypeRepresentations` field to our metadata, similar to aggregate functions and comparison operators
2. We already query for the enum types and variants, now we also return those from sql to rust and into the metadata.
3. For each scalar type, we try to look for their name in type reps, and return it if we find it.

### DIFF
--- a/crates/connectors/ndc-postgres/src/schema.rs
+++ b/crates/connectors/ndc-postgres/src/schema.rs
@@ -31,7 +31,27 @@ pub async fn get_schema(
             (
                 scalar_type.0.clone(),
                 models::ScalarType {
-                    representation: None,
+                    representation: metadata.type_representations.0.get(scalar_type).map(
+                        |type_representation| match type_representation {
+                            metadata::TypeRepresentation::Boolean => {
+                                models::TypeRepresentation::Boolean
+                            }
+                            metadata::TypeRepresentation::Integer => {
+                                models::TypeRepresentation::Integer
+                            }
+                            metadata::TypeRepresentation::Number => {
+                                models::TypeRepresentation::Number
+                            }
+                            metadata::TypeRepresentation::String => {
+                                models::TypeRepresentation::String
+                            }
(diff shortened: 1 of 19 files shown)


crates/connectors/ndc-postgres/src/schema.rs
```diff
@@ -479,7 +499,7 @@ fn make_procedure_type(
     scalar_types
         .entry("int4".to_string())
         .or_insert(models::ScalarType {
-            representation: None,
+            representation: Some(models::TypeRepresentation::Integer),
```
plcplc, line 502: This is really something of an edge case. I wonder if we actually need to register `int4`, as in, I wonder if there actually is a functional difference between `"scalarTypes":{"int4": {"aggregateFunctions": {}, "comparisonOperators": {}}` and `"scalarTypes": {}`.

At any rate, it is a curious detail that this edge case gives a different result than the common case where `int4` is defined, but we don't yet assign type representations to anything but enums.

plcplc (approved):
Looks good to me~

crates/connectors/ndc-postgres/src/schema.rs
soupi: Yes, we definitely need it. I think I added it because a test failed. Try and see :)

The comments about this are not visible in this diff by default, adding it here:

```
    // If int4 doesn't exist anywhere else in the schema, we need to add it here. However, a user
    // can't filter or aggregate based on the affected rows of a procedure, so we don't need to add
    // any aggregate functions or comparison operators. However, if int4 exists elsewhere in the
    // schema and has already been added, it will also already contain these functions and
    // operators.
```

PR #398, opened 2024-04-01
Support type representations in general

plcplc:
### What

This PR adds support for type representations for base types and domain types, building on top of the existing support for type representation hinting of enum types.

Note that the wide numeric types default to numeric representation, see the comment in the code. This will be relayed in the user docs in a follow-up PR.

### How

We introduce a new field `introspectionOptions.typeRepresentations` containing defaults for common built-in base types.

### DIFF
--- a/crates/configuration/src/version3/mod.rs
+++ b/crates/configuration/src/version3/mod.rs
@@ -104,7 +104,10 @@ pub async fn introspect(
             &args
                 .introspection_options
                 .introspect_prefix_function_comparison_operators,
-        );
+        )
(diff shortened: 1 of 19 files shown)


crates/configuration/src/version3/mod.rs
```diff
@@ -193,6 +195,103 @@ pub async fn introspect(
     })
 }
 
+fn base_type_representations() -> database::TypeRepresentations {
```
soupi, line 198: Let's add a comment saying what this is please.

soupi (approved):
Nice!

PR #404, opened 2024-04-03
Support nested fields of object type

plcplc:
### What

This PR adds support for selecting nested fields that arise from composite objects.

The case for nested fields which are arrays will be treated in a followup PR.

### How

* Handling of field selection is now recursive and lives in its own function.
* A dedicated case handling nested fields has been added. Each field at each layer of nesting adds another lateral join

The SQL generated is of the form
```
///   SELECT
///     coalesce(json_agg(row_to_json("%6_rows")), '[]') AS "rows"
///   FROM
///     (
///       SELECT
///         "%3_nested_fields_collect"."collected" AS "result"
///       FROM
///         <current table> AS "%0_<current table>"
///         LEFT OUTER JOIN LATERAL (
///           SELECT
///             ("%0_<current table>"."<composite column>").*
///         ) AS "%2_nested_field_bound" ON ('true')
///         LEFT OUTER JOIN LATERAL (
///           SELECT
///             row_to_json("%4_nested_fields") AS "collected"
///           FROM
///             (
///               SELECT
///                 "%2_nested_field_bound"."<nested column>" AS "<nested field alias>"
///             ) AS "%4_nested_fields"
///         ) AS "%3_nested_fields_collect" ON ('true')
///     ) AS "%6_rows"
```


crates/query-engine/sql/src/sql/helpers.rs
```diff
@@ -506,6 +522,29 @@ pub fn select_rows_as_json(
...
 
+/// Turn each row of a query result into a json object.
+///
+/// ```sql
+/// SELECT row_to_json(<table_alias>) AS <column_alias>
+/// FROM <query> as <table_alias>
+/// ```
+///
+/// - `row_to_json` takes a row and converts it to a json object.
+/// - `coalesce(<thing>, <otherwise>)` returns `<thing>` if it is not null, and `<otherwise>` if it is null.
```
soupi, line 533: This part is probably copy paste from the with_default version right? :)

crates/query-engine/translation/src/translation/query/root.rs
```diff
@@ -53,44 +61,67 @@ pub enum ReturnsFields {
     NoFieldsWereRequested,
 }
 
-/// Translate rows part of query to sql ast.
-pub fn translate_rows_query(
+struct JoinNestedFieldInfo {
```
soupi, line 64: Please add a docstring.

crates/query-engine/translation/src/translation/query/root.rs
```diff
@@ -53,44 +61,67 @@ pub enum ReturnsFields {
...
+    select: sql::ast::Select,
+    alias: sql::ast::TableAlias,
+}
+
+fn transalate_nested_field_joins(joins: Vec<JoinNestedFieldInfo>) -> Vec<sql::ast::Join> {
```
soupi, line 69: Please add a docstring.

crates/query-engine/translation/src/translation/query/root.rs
```diff
@@ -116,10 +147,159 @@ pub fn translate_rows_query(
...
+    state: &mut State,
+    // The table reference the column lives on
+    current_table: &TableNameAndReference,
+    // The column to extract nested fields from
+    current_column: &ColumnInfo,
+    field: models::NestedField,
+    join_nested_fields: &mut Vec<JoinNestedFieldInfo>,
+    join_relationship_fields: &mut Vec<relationships::JoinFieldInfo>,
+) -> Result<sql::ast::ColumnReference, Error> {
+    match field {
```
soupi, line 194: I would really appreciate some comments in-between some code chunks summarizing what they do. Rust can be quite verbose and it gives some direction to reading the code (and also easier to verify the intent matches the code).

crates/query-engine/translation/src/translation/query/root.rs
```diff
@@ -116,10 +147,159 @@ pub fn translate_rows_query(
...
+///             ) AS "%4_nested_fields"
+///         ) AS "%3_nested_fields_collect" ON ('true')
+///     ) AS "%6_rows"
+///
+fn translate_nested_field(
+    env: &Env,
+    state: &mut State,
+    // The table reference the column lives on
+    current_table: &TableNameAndReference,
+    // The column to extract nested fields from
```
soupi, line 188: Can these be `///`? Does that work?

soupi (changes requested):
I'm still in the process of reviewing and nothing seems to jump out, but I would appreciate more comments/docstrings on a few places, there are a couple of very large functions that do not have documentation.

plcplc:
@soupi: Thanks for the review. I've tried to improve the documentation :-)

crates/query-engine/translation/src/translation/query/root.rs
plcplc: It seems rust does not support `///` (or `#[doc()]` attributes) on function arguments. I've moved these to the main `///` block under an `# Aruguments` heading.

I'm fine either way myself.

crates/query-engine/translation/src/translation/query/root.rs
plcplc: Agree! I've added some commentary explaining how `translate_nested_field` goes about its business.

crates/query-engine/translation/src/translation/query/root.rs
plcplc: done

crates/query-engine/translation/src/translation/query/root.rs
plcplc: done.

crates/query-engine/sql/src/sql/helpers.rs
plcplc: Right - I've pruned it moer now. Thanks

soupi (approved):
I think this makes sense to me. Nice!

PR #407, opened 2024-04-05
bugfix: allow aliasing returning and affected_rows fields

soupi:
### What

It should be possible to alias the returning and affected_rows fields when running a mutation.
This is doubly important because this is what the cli/lsp/engine do by default for `affected_rows` :)

### How

Instead of having the alias satisfy the field name, we ask the Field to do that.
We traverse the result fields, find the ones named "returning" and "affected_rows", and use that, instead of trying to fetch by the alias name in the fields map.

### DIFF
--- a/crates/query-engine/translation/src/translation/mutation/translate.rs
+++ b/crates/query-engine/translation/src/translation/mutation/translate.rs
@@ -63,7 +63,7 @@ fn translate_mutation(
         alias: table_alias.clone(),
     };
 
-    let (aggregates, fields) = parse_procedure_fields(fields)?;
+    let (aggregates, (returning_alias, fields)) = parse_procedure_fields(fields)?;
 
     // define the query selecting from the native query,
     // selecting the affected_rows as aggregate and the fields.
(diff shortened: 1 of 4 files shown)


crates/query-engine/translation/src/translation/mutation/translate.rs
```diff
@@ -280,49 +280,52 @@ pub fn parse_procedure_fields(
...
-                                        Err(Error::UnexpectedStructure(
-                                            "multi-dimensional array in 'returning' clause"
-                                                .to_string(),
-                                        ))?
+            let mut affected_rows = None;
+            let mut returning = ("returning".to_string(), None);
+
+            for (alias, field) in fields {
+                match field {
+                    models::Field::Column { column, fields: _ } if column == "affected_rows" => {
```
soupi, line 294: we check the field name instead of the alias now.

crates/tests/tests-common/goldenfiles/mutations/delete_playlist_track.json
```diff
@@ -9,12 +9,11 @@
       "fields": {
         "type": "object",
         "fields": {
-          "affected_rows": {
+          "affectedRows": {
```
soupi, line 12: now we can give a different alias.

SamirTalwar (approved):

PR #408, opened 2024-04-05
ndc-spec 0.1.2

soupi:
### What

Support ndc-spec v0.1.2.
This includes changing the type representations of some types.

### How

The type representation has changed. We try to keep this change backwards compatible by mapping deprecated variants (Number, Integer) to new variants, and update types to more specific reps that didn't exist before.

We also have a problem with representing int8 now. ndc-spec v0.1.2 defines Int64 to be represented as a json string, which is not what we do. So we either have to lie (say it's an Int32 when it isn't) or break compatibility and cast int8 to string.
In this PR I did the former, and I'll introduce the latter change next.


crates/connectors/ndc-postgres/src/schema.rs
```diff
@@ -626,3 +614,34 @@ fn make_procedure_type(
         },
     }
 }
+
+fn map_type_representation(
+    type_representation: &metadata::TypeRepresentation,
+) -> models::TypeRepresentation {
+    match type_representation {
+        metadata::TypeRepresentation::Integer => models::TypeRepresentation::Int32,
+        metadata::TypeRepresentation::Number => models::TypeRepresentation::Float64,
```
soupi, line 623: Stop gap until we remove these.

SamirTalwar (approved):

SamirTalwar:
Dismissing my review because I just realized that the ndc-sdk commit points to a branch. Let's wait until that's merged and update the commit.

crates/connectors/ndc-postgres/src/schema.rs
```diff
@@ -626,3 +614,34 @@ fn make_procedure_type(
         },
     }
 }
+
+/// Map our local type representation to ndc-spec type representation.
+fn map_type_representation(
```
danieljharvey, line 619: I'm not sure we should be changing the outputted type representations like this without the user providing us an opt-in config flag to do so.

soupi: This part is just translating our own TypeRepresentation type to ndc-spec TypeRepresentation type. Could you please elaborate on what do you mean? I'm not sure I understand.

danieljharvey: I was under the impression that our types should return the same `ndc-spec` `TypeRepresentation` as they did already before this change. Where this is not possible, because of removed types, we should return the more general `json`. Otherwise we're potentially changing behaviour without the user opting into it.

static/cockroach/v3-chinook-ndc-metadata/configuration.json
```diff
@@ -2803,19 +2803,19 @@
...
-      "int2": "integer",
-      "int4": "integer",
-      "int8": "integer",
-      "numeric": "number",
+      "date": "date",
+      "float4": "float32",
+      "float8": "float64",
+      "int2": "int16",
+      "int4": "int32",
+      "int8": "json",
```
soupi, line 2811: These are the changes to the configuration. Number and String are deprecated in ndc-spec. We will still accept these for backwards compatibility reasons, but we will output the new representations. @danieljharvey

crates/tests/databases-tests/src/postgres/snapshots/databases_tests__postgres__schema_tests__schema_test__get_schema.snap
```diff
@@ -752,7 +752,7 @@ expression: result
     },
     "even_number": {
       "representation": {
-        "type": "integer"
+        "type": "int32"
```
soupi, line 755: These are the changes to the ndc schema.

crates/connectors/ndc-postgres/src/schema.rs
danieljharvey: What will a version of engine that does not know about these types do when it receives them?

soupi: This is an ndc-schema concept that can be used by cli/lsp to generate engine objects, so I don't think this will be used directly in the engine. The engine already uses v0.1.2 of ndc-spec though: https://github.com/hasura/v3-engine/pull/443/files#diff-5e05c14aa4760c15d75bc66970581abe17a82eef4770a0cd9861bd40ce65934bR27

danieljharvey: Ahh OK, makes sense. This all seems sensible then, thank you for indulging my questions.

danieljharvey (approved):

Cargo.toml
```diff
@@ -47,6 +47,6 @@ unreadable_literal = "allow"
 unused_async = "allow"
 
 [workspace.dependencies]
-ndc-client = { git = "https://github.com/hasura/ndc-spec.git", tag = "v0.1.1" }
-ndc-sdk = { git = "https://github.com/hasura/ndc-sdk-rs.git", rev = "7b56fac3aba2bc6533d3163111377fd5fbeb3011" }
-ndc-test = { git = "https://github.com/hasura/ndc-spec.git", tag = "v0.1.1" }
+ndc-models = { git = "https://github.com/hasura/ndc-spec.git", tag = "v0.1.2" }
+ndc-sdk = { git = "https://github.com/hasura/ndc-sdk-rs.git", rev = "7409334" }
```
SamirTalwar, line 51: I would prefer to use the full SHA here because Cargo thinks two dependencies pointing to the _same_ revision but using a _different_ prefix length are actually different dependencies. This causes serious issues when using ndc-postgres as a dependency as it brings in ndc-sdk with the short ref, so everything else needs to use the exact same prefix length.

soupi: Good to know. Will fix.

PR #409, opened 2024-04-05
 Support nested fields of array type

plcplc:
### What

This PR adds support for selecting nested fields that arise from composite arrays.

### How

This builds on the earlier work for nested field selection support for objects.

The main change is that the lateral joins that project the nested field have to be nested rather than adjacent as well, since there are multiple rows in the case of arrays.

The generated SQL is now of the form:
```
LEFT OUTER JOIN LATERAL (
  SELECT
    <collect_expression> AS "collected"
  FROM
    (
      < select of translate_fields(fields, nested_field_binding_alias, ... )
        which includes joins from recursive calls>
       FROM
         (
            SELECT
             (<field_binding_expression>).*
         ) AS <nested_field_binding> ON ('true')
     ) AS <nested_fields>
) AS <nested_fields_collect> ON ('true')
```


crates/query-engine/translation/tests/goldenfiles/select_nested_column_complex/request.json.bak
soupi, line 1: bak?

crates/query-engine/translation/src/translation/query/root.rs
soupi, line 1: suggestion: move translate fields related functionality into it's own module named `fields.rs`. There' currently a lot of code here that is dedicated to a specific task, it makes sense it'll have it's own module :)

crates/query-engine/translation/tests/goldenfiles/select_nested_column_complex/request.json.bak
plcplc: Oops, this was staged by accident. I have removed it.

crates/query-engine/translation/src/translation/query/root.rs
plcplc: I've moved it to its own file now.

crates/query-engine/translation/src/translation/query/fields.rs
```diff
@@ -0,0 +1,320 @@
...
+        .joins
+        .extend(transalate_nested_field_joins(nested_field_joins));
+
+    Ok(select)
+}
+
+/// Translate a nested field selection.
+///
+/// Nested fields are different from relationships in that the value of a nested field is already
+/// avaliable on the current table as a column of composite type.
```
danieljharvey, line 122: (sorry)

```suggestion
/// available on the current table as a column of composite type.
```

danieljharvey (approved):
Thank you for going into such detail with the comments and the examples.

PR #416, opened 2024-04-09
cast int64 and numeric to string by default

soupi:
### What

We want to be able to stringify fields of types `int8` and `numeric` in the query response so that we have better interoperability with javascript clients.

This PR changes the default type representation of `int8` and `numeric` types (maintaining the existing type representation if one exists) and stringify numbers.

This option can be change via the configuration by editing the mapping in the `metadata.typeRepresentations`:

```diff
  { ...
    "metadata": {
      ...
      "typeRepresentations": {
        "Phone": "string",
        "CardSuit": {
          "enum": ["hearts", "clubs", "diamonds", "spades"]
        },
        ...
-       "int8": "int64",
+       "int8": "int64AsString",
-       "numeric": "bigDecimal",
+       "numeric": "bigDecimalAsString",
        ...
      }
      ...
    }
  }
```

### How

1. We introduce new type representations: `Int64AsString`, `BigDecimalAsString`.
2. Make sure we use the existing type representation configuration and override the default if it exists.
3. In field selection after getting an expression, we cast it if the type representation requires it. (For arrays, we cast to `cast_type[]`)
4. For composite types we unpack them if the don't specify fields, and cast required fields recursively.


crates/query-engine/translation/src/translation/helpers.rs
```diff
@@ -89,13 +90,17 @@ pub enum CollectionInfo<'env> {
 }
 
 #[derive(Debug)]
-/// Metadata information about a specific collection.
-pub enum CompositeTypeInfo<'env> {
+/// Metadata information about a specific collection or composite type.
+pub enum CollectionOrCompositeTypeInfo<'env> {
```
plcplc, line 94: The reason I called it `CompositeTypeInfo` is that the fact that it represents a type is the interesting thing. Whether this type comes from a collection or a separate composite type in the metadata is only interesting as an implementation detail to lazily query the columns of the type.

soupi: I agree in theory, but in practice they are represented a bit differently in data.

plcplc: Oh, do you mean whether a type originates from a table vs a distinct composite type has bearing on how its type representations are processed? (I haven't read through the PR completely yet)

soupi: I'll let you finish and get your feedback :)

crates/query-engine/translation/src/translation/query/fields.rs
```diff
@@ -318,3 +323,135 @@ fn translate_nested_field(
...
+                let composite_type = env.lookup_composite_type(composite_type)?;
+                let mut fields = indexmap!();
+                for (name, field_info) in composite_type.info.fields.iter() {
+                    fields.insert(
+                        name.to_string(),
+                        models::Field::Column {
+                            column: field_info.name.clone(),
+                            fields: None,
+                        },
+                    );
```
plcplc, line 375: Ah, I think this is the root of our CompositeTypeInfo/CollectionOrCompositeTypeInfo-related misunderstanding.

It's perfectly possible for our `composite_type` bound in `Type::CompositeType(ref composite_type)` to come from both tables _and_ composite types.

crates/query-engine/translation/src/translation/query/fields.rs
```diff
@@ -318,3 +323,135 @@ fn translate_nested_field(
...
+fn unpack_and_wrap_fields(
+    env: &Env,
+    state: &mut State,
+    current_table: &TableNameAndReference,
+    join_relationship_fields: &mut Vec<relationships::JoinFieldInfo>,
+    column: &str,
+    alias: sql::ast::ColumnAlias,
+    type_info: &CollectionOrCompositeTypeInfo<'_>,
+    nested_field_joins: &mut Vec<JoinNestedFieldInfo>,
+) -> Result<(sql::ast::ColumnAlias, sql::ast::Expression), Error> {
```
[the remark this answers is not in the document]
soupi: I actually pulled it out because that function has gotten quite big and found this to be more digestible, but we can talk about this if you really prefer it to be inlined.

crates/query-engine/translation/src/translation/helpers.rs
```diff
@@ -202,6 +223,23 @@ impl<'request> Env<'request> {
             })
     }
 
+    /// Lookup type representation of a type.
+    pub fn lookup_type_representation(
+        &self,
+        typ: &metadata::Type,
```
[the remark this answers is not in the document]
soupi: Makes sense. Perhaps we can add a method on type `scalar_type() : Type -> Option<ScalarType>` that we can use on the caller side and it will make this more straightforward.

crates/connectors/ndc-postgres/src/schema.rs
```diff
@@ -626,8 +626,14 @@ fn map_type_representation(
...
-        metadata::TypeRepresentation::Int64 => models::TypeRepresentation::Int64,
-        metadata::TypeRepresentation::BigDecimal => models::TypeRepresentation::BigDecimal,
+        // Int64 returns a number.
+        metadata::TypeRepresentation::Int64 => models::TypeRepresentation::JSON,
+        // Int64AsString returns a string.
+        metadata::TypeRepresentation::Int64AsString => models::TypeRepresentation::Int64,
+        // BigDecimal returns a number.
+        metadata::TypeRepresentation::BigDecimal => models::TypeRepresentation::JSON,
+        // BigDecimalAsString returns a string.
+        metadata::TypeRepresentation::BigDecimalAsString => models::TypeRepresentation::BigDecimal,
```
soupi, line 636: This is the mapping between our internal type representations and ndc-spec type representations.

crates/tests/databases-tests/src/postgres/snapshots/databases_tests__postgres__query_tests__basic__select_array_composite_field.snap
soupi, line 1: A test of array of composite types.

crates/query-engine/translation/src/translation/helpers.rs
```diff
@@ -202,6 +203,18 @@ impl<'request> Env<'request> {
...
+    /// Lookup type representation of a type.
+    pub fn lookup_type_representation(
+        &self,
+        scalar_type: &metadata::ScalarType,
+    ) -> Option<metadata::TypeRepresentation> {
+        self.metadata
+            .type_representations
+            .0
+            .get(scalar_type)
+            .cloned()
```
SamirTalwar, line 215: Stupid nitpick: we only use this as a reference so we could probably return one and not need to clone it.

This is stupid because cloning it will be cheap.

crates/query-engine/translation/src/translation/query/fields.rs
```diff
@@ -318,3 +323,201 @@ fn translate_nested_field(
...
+            )?;
+
+            nested_field_joins.push(nested_field_join);
+
+            Ok((
+                alias,
+                sql::ast::Expression::ColumnReference(nested_column_reference),
+            ))
+        }
+        // TODO: Arrays of composite types are not handled yet.
```
SamirTalwar, line 384: I am confused; it looks like they are. Should this say "Arrays of arrays are not handled yet"?

crates/query-engine/translation/src/translation/query/fields.rs
```diff
@@ -318,3 +323,201 @@ fn translate_nested_field(
...
+        TypeRepresentation::Timestamp => None,
+        TypeRepresentation::Timestamptz => None,
+        TypeRepresentation::Time => None,
+        TypeRepresentation::Timetz => None,
+        TypeRepresentation::Date => None,
+        TypeRepresentation::UUID => None,
+        TypeRepresentation::Geography => None,
+        TypeRepresentation::Geometry => None,
+        TypeRepresentation::Json => None,
+        TypeRepresentation::Enum(_) => None,
```
SamirTalwar, line 503: I appreciate you listing these out rather than using a wildcard.

You can use `|` to join them, though (there's a pedantic warning for this but we haven't enabled it).

SamirTalwar (approved):

crates/query-engine/translation/src/translation/query/fields.rs
soupi: Thanks! stale comment.

crates/query-engine/translation/src/translation/helpers.rs
soupi: smart. thanks!

crates/query-engine/translation/src/translation/query/fields.rs
soupi: cool. Thanks!

PR #419, opened 2024-04-11
Separate business and exchange types

plcplc:
### What

We want to be able to confidently evolve the configuration format without risk of accidentally changing the behavior of previous versions.

This PR has no user-observable change in behavior.

Introducing a new major version of the configuration format is now a much more well-defined task, which should consist most of just duplicating the entire `version3` module hierarchy and then introducing the new changes in the duplicated copy.

### How

In order to achieve this we duplicate the complete metadata type hierarchy into configuration crate and introduce functions that convert between the now separate exchange types and business types, and ensure that all logic that has to do with serialization only lives with the exchange types (the version types)

### DIFF
--- a/crates/configuration/src/configuration.rs
+++ b/crates/configuration/src/configuration.rs
@@ -4,13 +4,12 @@ use std::path::Path;
 
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
-use tokio::fs;
 
 use query_engine_metadata::metadata;
 
 use crate::environment::Environment;
 use crate::error::Error;
-use crate::values::{ConnectionUri, IsolationLevel, PoolSettings, Secret};
+use crate::values::{IsolationLevel, PoolSettings};
 use crate::version3;
 
 pub const CONFIGURATION_FILENAME: &str = "configuration.json";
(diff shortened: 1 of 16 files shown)


crates/configuration/src/configuration.rs
```diff
@@ -59,45 +58,6 @@ pub async fn parse_configuration(
...
-                }),
-        }?;
-    Ok(Configuration {
-        metadata: configuration.metadata,
-        pool_settings: configuration.connection_settings.pool_settings,
-        connection_uri,
-        isolation_level: configuration.connection_settings.isolation_level,
-        mutations_version: configuration.mutations_version,
-    })
+    // Try parsing each supported version in turn
```
SamirTalwar, line 61: I see you are predicting the future.

crates/configuration/src/version3/mod.rs
SamirTalwar, line 1: This is a very long way to say `coerce`.

SamirTalwar (approved):

crates/query-engine/metadata/src/metadata/mod.rs
```diff
@@ -4,27 +4,20 @@ pub mod database;
...
     pub composite_types: CompositeTypes,
-    #[serde(default)]
     pub native_queries: NativeQueries,
-    #[serde(default)]
     pub aggregate_functions: AggregateFunctions,
-    #[serde(default)]
     pub comparison_operators: ComparisonOperators,
-    #[serde(default)]
     pub type_representations: TypeRepresentations,
+    pub occurring_scalar_types: BTreeSet<ScalarType>,
```
soupi, line 22: This seems new. Who is this?

crates/configuration/src/version3/mod.rs
soupi: I think it would be nice to split up this module to a few topics and add docstring to each function. Perhaps a `parsing` directory with modules for each field in Metadata?

soupi (commented):
I think this is very nice! I like the separation between the types and the parsing. I think there's an opportunity for an easier to follow hierarchy and a few more docstrings. Also had a question about the introduction of a new field to Metadata.

crates/configuration/src/configuration.rs
plcplc: 🔮 Indeed

crates/configuration/src/version3/mod.rs
plcplc: Yeah, if it were haskell :-)

But at the very least what this gives you are concrete slots where you can make changes. If this PR had just been `convert_metadata = coerce` then the next guy wanting to change, say `ScalarType` to something non-coercible would have to make all these definitions anyway. :-)

plcplc: I think it would be nice to have most of these functions in a separate module yeah.
But I think that as long as most are simple 1:1 mappings it's nicer to have them in the same module for the locality it brings.

One thing I think is that a separate, named function for each type is superior to any other solution I can think of. We could have used e.g. `From`, and that would have worked while business and exchange types are congruent but breaks down once we want them not to be. And once one type conversion function needs access to specific context a trait has no advantage over a plain old function.

PR #420, opened 2024-04-11
include additional schemas for types by default

soupi:
### What

Include a couple more unqualified schemas for types and procedures defaults.
See follow-up ticket for next task: https://hasurahq.atlassian.net/browse/PG-82

### How

We expose the new fields in the defaults. These are exposed in the initialize phase so if a user already has a list, this won't change it.


crates/configuration/src/version3/options.rs
```diff
@@ -73,6 +73,8 @@ fn default_unqualified_schemas_for_types_and_procedures() -> Vec<String> {
         "public".to_string(),
         "pg_catalog".to_string(),
         "tiger".to_string(),
+        "auth".to_string(),
+        "pgsodium".to_string(),
```
SamirTalwar, line 77: Pleae add comments to each of these explaining which extension requires it.

plcplc (approved):

PR #421, opened 2024-04-15
remove transactions in queries

soupi:
### What

Running queries in transactions require that we run 3 statements: begin, the query, commit.
We use transactions to enforce a `READ ONLY` setting so that native queries which run mutations will be rejected.
However, each roundtrip takes time. If we can avoid it, it would improve the performance of queries.

In the PR we use the fact the mutating sql statements can only occur in a CTE at the top-level, and wrap native queries in another layer of CTE, thereby enforce the read only requirement without using transactions.

### How

1. We remove the `begin` and `commit` statements before/after a query.
2. We wrap each query in another CTE of the form `WITH <native_query_name> as (WITH <alias> as (<native_query>) SELECT * FROM CTE as <alias>) SELECT ...`
3. We extract the global table alias index from the state and use it in wrapping the CTEs.
4. We wrap ctes in the caller side because this code is used for mutations as well, for which we don't want to apply this hack.


crates/query-engine/sql/src/sql/execution_plan.rs
```diff
@@ -50,22 +50,18 @@ pub fn explain_to_sql(explain: &sql::ast::Explain) -> sql::string::SQL {
...
     variables: Option<Vec<BTreeMap<String, serde_json::Value>>>,
     root_field: String,
     query: sql::ast::Select,
 ) -> ExecutionPlan<Query> {
     ExecutionPlan {
-        pre: sql::helpers::begin(
-            isolation_level,
-            sql::ast::transaction::TransactionMode::ReadOnly,
-        ),
+        pre: vec![],
```
soupi, line 58: No more begin and commit for queries.

crates/query-engine/translation/src/translation/helpers.rs
```diff
@@ -315,26 +315,17 @@ impl State {
...
-    }
-
-    /// increment the table index and return the current one.
-    fn next_global_table_index(&mut self) -> TableAliasIndex {
-        let TableAliasIndex(index) = self.global_table_index;
-        self.global_table_index = TableAliasIndex(index + 1);
-        TableAliasIndex(index)
+    /// Fetch the tracked native queries used in the query plan and their table alias,
+    /// and the global table index.
+    pub fn get_native_queries_and_global_index(self) -> (Vec<NativeQueryInfo>, TableAliasIndex) {
```
soupi, line 320: We break down the state so we can consume the native queries when processing native queries, and use the global table alias index outside of it when inventing names for the wrapping ctes.

crates/query-engine/translation/src/translation/mutation/translate.rs
```diff
@@ -257,7 +257,8 @@ fn translate_native_query(
 
     // add the procedure native query definition is a with clause.
     select.with = sql::ast::With {
-        common_table_expressions: crate::translation::query::native_queries::translate(env, state)?,
+        common_table_expressions: crate::translation::query::native_queries::translate(env, state)?
```
soupi, line 260: mutations do need to run native queries at the top-level, because they are mutations. So we don't wrap the ctes here.

crates/query-engine/translation/src/translation/query/mod.rs
```diff
@@ -66,7 +65,13 @@ pub fn translate(
...
         // native queries if there are any
         sql::ast::With {
-            common_table_expressions: native_queries::translate(&env, state)?,
+            common_table_expressions: {
+                let (ctes, mut global_table_index) = native_queries::translate(&env, state)?;
+                // wrap ctes in another cte to guard against mutations in queries
+                ctes.into_iter()
+                    .map(|cte| native_queries::wrap_cte_in_cte(&mut global_table_index, cte))
+                    .collect()
+            },
```
soupi, line 74: here we wrap.

crates/query-engine/translation/src/translation/query/native_queries.rs
```diff
@@ -72,5 +75,38 @@ pub fn translate(env: &Env, state: State) -> Result<Vec<sql::ast::CommonTableExp
         });
     }
 
-    Ok(ctes)
+    Ok((ctes, global_table_index))
```
soupi, line 78: We return the global table index here as well so it can be used when wrapping ctes, but the native queries with the state is consumed.

crates/query-engine/translation/src/translation/query/native_queries.rs
```diff
@@ -72,5 +75,38 @@ pub fn translate(env: &Env, state: State) -> Result<Vec<sql::ast::CommonTableExp
...
+        };
+        select
+    });
+
+    // wrap in another CTE.
+    sql::ast::CommonTableExpression {
+        alias: outer_cte_alias,
+        column_names: None,
+        select: nested_cte_select,
+    }
```
soupi, line 111: how to wrap ctes.

crates/query-engine/translation/tests/snapshots/tests__it_select_where_in_variable.snap
```diff
@@ -15,16 +12,22 @@ FROM
...
-            3 as three,
-            array_agg(arr.series) AS series
-          FROM
-            (
-              SELECT
-                generate_series(
-                  cast(
-                    (
-                      ("%0_%variables_table"."%variables" -> $2) #>> cast(ARRAY [] as text[])) as int4),cast((("%0_%variables_table"."%variables" -> $3) #>> cast(ARRAY [] as text[])) as int4)) AS series) AS arr
+          WITH "%9_NATIVE_QUERY_array_series" AS (
```
soupi, line 15: an example of a wrapping.

crates/query-engine/translation/tests/snapshots/tests__it_aggregate_function_albums.snap
soupi, line 1: begin and commit no more.

crates/tests/databases-tests/src/postgres/snapshots/databases_tests__postgres__query_tests__negative__mutation_native_query_in_query.snap
soupi, line 1: Test that this works.

SamirTalwar: This is what I wanted to see. Thank you!

SamirTalwar (approved):

PR #424, opened 2024-04-16
cli: remove the 'Caused by' section in the error message

soupi:
### What

https://hasurahq.atlassian.net/browse/V3DX-1053

Before:

```
$ CONNECTION_URI="postgresql://re" cargo run --bin ndc-postgres-cli -- --context='/tmp/baba' update
Error: error communicating with database: failed to lookup address information: Name or service not known

Caused by:
    failed to lookup address information: Name or service not known
```

After:

```
$ CONNECTION_URI="postgresql://re" cargo run --bin ndc-postgres-cli -- --context='/tmp/baba' update
ERROR: error communicating with database: failed to lookup address information: Name or service not known
```

### How

Instead of having a `main() -> anyhow::Result<()>`, we explicitly handle printing the error.
https://docs.rs/anyhow/1.0.69/anyhow/struct.Error.html#display-representations


crates/cli/src/main.rs
```diff
@@ -31,10 +31,19 @@ pub struct Args {
     pub subcommand: Command,
 }
 
+#[tokio::main]
+pub async fn main() {
```
SamirTalwar, line 35: Perhaps return [`ExitCode`](https://doc.rust-lang.org/std/process/struct.ExitCode.html) here rather than calling `std::process::exit`.

SamirTalwar (approved):

PR #429, opened 2024-04-17
Use idiomatic iterator patterns.

SamirTalwar:
### What

Just fixing a few lints for fun.

### How

Mostly, deleting the exclusions and following the guidance. `clippy --fix` did most of the work.


soupi (approved):
cool

PR #430, opened 2024-04-18
Update broken metadata, and test that it is kept up to date

plcplc:
### What

The `just generate-metadata` command updated the "broken" test metadata on `main`. This PR introduces a test that ensures this is kept up to date, such that updates to this metadata doesn't lag behind the changes that actually introduced them.


soupi (approved):

SamirTalwar:
Looks like you need to start supplying separate metadata for older versions of PostgreSQL.

plcplc:
> Looks like you need to start supplying separate metadata for older versions of PostgreSQL.

.. Or only run this test on the latest Postgres, which is what the adjacent is-the-config-idempotent tests do as well.

soupi:
If this is because the `text` field, we can remove that column and do `RETURNING "ArtistId"` instead.

PR #431, opened 2024-04-18
Add Schema support for Scalar types in SQL AST

plcplc:
### What

This PR is one step on the way to supporting scalar types living in non-search-path schemas.

It adds fields to the representation of a scalar type name at the SQL AST level, indicating the schema the type is defined in and whether the type is an array type.

### DIFF
--- a/crates/query-engine/metadata/src/metadata/database.rs
+++ b/crates/query-engine/metadata/src/metadata/database.rs
@@ -9,14 +9,13 @@ pub struct CompositeTypes(pub BTreeMap<String, CompositeType>);
 
 /// A Scalar Type.
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
-
-pub struct ScalarType(pub String);
+pub struct ScalarTypeName(pub String);
 
 /// The type of values that a column, field, or argument may take.
 #[derive(Debug, Clone, PartialEq, Eq)]
 
 pub enum Type {
-    ScalarType(ScalarType),
+    ScalarType(ScalarTypeName),
     CompositeType(String),
     ArrayType(Box<Type>),
 }
@@ -46,7 +45,7 @@ pub struct FieldInfo {
 /// Not all of these are supported for every type.
 #[derive(Debug, Clone, PartialEq, Eq, Default)]
 
-pub struct ComparisonOperators(pub BTreeMap<ScalarType, BTreeMap<String, ComparisonOperator>>);
+pub struct ComparisonOperators(pub BTreeMap<ScalarTypeName, BTreeMap<String, ComparisonOperator>>);
 
 /// Represents a postgres binary comparison operator
 #[derive(Debug, Clone, PartialEq, Eq)]
@@ -54,7 +53,7 @@ pub struct ComparisonOperators(pub BTreeMap<ScalarType, BTreeMap<String, Compari
 pub struct ComparisonOperator {
     pub operator_name: String,
     pub operator_kind: OperatorKind,
-    pub argument_type: ScalarType,
+    pub argument_type: ScalarTypeName,
 
     pub is_infix: bool,
 }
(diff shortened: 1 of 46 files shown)


crates/query-engine/metadata/src/metadata/database.rs
```diff
@@ -9,14 +9,13 @@ pub struct CompositeTypes(pub BTreeMap<String, CompositeType>);
 
 /// A Scalar Type.
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
-
-pub struct ScalarType(pub String);
+pub struct ScalarTypeName(pub String);
```
plcplc, line 12: In order to be precise in our terminology, we now take care to point out that we are referring explicitly to the name of a scalar type.

This distinction will become more relevant in subsequent PRs when we distinguish between the scalar type name at the ndc level vs on the database.

crates/query-engine/sql/src/sql/helpers.rs
```diff
@@ -624,14 +624,14 @@ pub fn select_row_as_json_with_default(
 /// ```
 pub fn from_variables(alias: TableAlias) -> From {
     let expression = Expression::Value(Value::Variable(VARIABLES_OBJECT_PLACEHOLDER.to_string()));
-    let columns: Vec<(ColumnAlias, ScalarType)> = vec![
+    let columns: Vec<(ColumnAlias, ScalarTypeName)> = vec![
         (
             make_column_alias(VARIABLE_ORDER_FIELD.to_string()),
-            ScalarType("int".to_string()),
+            ScalarTypeName::new_unqualified("int4"),
```
plcplc, line 630: The change `int` -> `int4` is due to the second quirk of type name qouting.

crates/query-engine/translation/src/translation/query/values.rs
```diff
@@ -51,27 +51,31 @@ pub fn translate_json_value(
...
-            sql::ast::ScalarType(scalar_type + "[]")
+            // This will collapse nested arrays. This is fine since it matches the behavior of
+            // Postgres where these are unsupported anyway.
+            let mut scalar_type = type_to_ast_scalar_type(t);
+            scalar_type.is_array = true;
+            scalar_type
+        }
+        query_engine_metadata::metadata::Type::ScalarType(t) => {
+            // TODO: This will need access to a mapping between ndc-type names and db type names
+            sql::ast::ScalarTypeName::new_unqualified(&t.0)
```
plcplc, line 73: `ast::ScalarTypeName` now supports schemas, but we do not track them yet from introspection.

This will change in subsequent PRs.

crates/query-engine/translation/tests/snapshots/tests__it_select_where_in_variable.snap
```diff
@@ -9,7 +9,7 @@ FROM
     SELECT
       row_to_json("%5_universe") AS "universe"
     FROM
-      jsonb_to_recordset($1) AS "%0_%variables_table"("%variable_order" int, "%variables" jsonb)
+      jsonb_to_recordset($1) AS "%0_%variables_table"("%variable_order" "int4", "%variables" "jsonb")
```
plcplc, line 12: Most of the changes in this PR are knock-on effects on test snapshots from type name identifiers now being quoted.

SamirTalwar (approved):
I see a lot of extra quoting, which I like very much.

crates/query-engine/sql/src/sql/ast.rs
```diff
@@ -299,9 +299,35 @@ pub enum Value {
...
+///
+/// * Quoting of type name identifiers is only supported for the actual type names recorded in
+/// `pg_type`, and _not_ the SQL standard type names. This means that `character varying` is an
+/// acceptable type name, but `"character varying"` is _not_ (Unless of course you do `CREATE TYPE
+/// "character varying" AS (..)`. Spicy).
+///
 #[derive(Debug, Clone, PartialEq)]
-pub struct ScalarType(pub String);
+pub struct ScalarTypeName {
+    pub type_name: String,
```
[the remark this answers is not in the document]
plcplc: Yeah, there is some `.clone()`ing and `.to_string()`ing going on because of this, especially for `CompositeTypeName` in the next PR. I guess I just didn't know how to solve this in Rust yet.

PR #432, opened 2024-04-18
Add schema names to composite types.

plcplc:
### What

This PR adds schema names to composite types in metadata. Schema names are not introspected yet, but this will happen in a subsequent PR.

Also translate composite types into sql::ast::ScalarTypeName using their
name and schema rather than the ndc-schema object type name, which had
incidentally been the same as the (unqualified) type name.

This means we need access to the `Env` when translating values/variables of
composite type. Most of the files changed in this PR is a result from
knock-on effects of this.

This PR is a follow-up to https://github.com/hasura/ndc-postgres/pull/431.

### DIFF
--- a/crates/query-engine/metadata/src/metadata/database.rs
+++ b/crates/query-engine/metadata/src/metadata/database.rs
@@ -22,14 +22,20 @@ pub enum Type {
 #[derive(Debug, Clone, PartialEq, Eq, Default)]
 pub struct CompositeTypes(pub BTreeMap<CompositeTypeName, CompositeType>);
 
+impl CompositeTypes {
+    pub fn empty() -> Self {
+        CompositeTypes(BTreeMap::new())
+    }
+}
+
 /// Information about a composite type. These are very similar to tables, but with the crucial
 /// difference that composite types do not support constraints (such as NOT NULL).
 #[derive(Debug, Clone, PartialEq, Eq)]
 
 pub struct CompositeType {
     pub name: String,
+    pub schema_name: Option<String>,
     pub fields: BTreeMap<String, FieldInfo>,
-
     pub description: Option<String>,
 }
 
@@ -48,6 +54,12 @@ pub struct FieldInfo {
 #[derive(Debug, Clone, PartialEq, Eq, Default)]
 
 pub struct ComparisonOperators(pub BTreeMap<ScalarTypeName, BTreeMap<String, ComparisonOperator>>);
+
+impl ComparisonOperators {
+    pub fn empty() -> Self {
+        ComparisonOperators(BTreeMap::new())
+    }
+}
 
 /// Represents a postgres binary comparison operator
 #[derive(Debug, Clone, PartialEq, Eq)]
@@ -73,6 +85,12 @@ pub enum OperatorKind {
 #[derive(Debug, Clone, PartialEq, Eq, Default)]
 
 pub struct TablesInfo(pub BTreeMap<String, TableInfo>);
+
+impl TablesInfo {
+    pub fn empty() -> Self {
+        TablesInfo(BTreeMap::new())
+    }
+}
 
 /// Information about a database table (or any other kind of relation).
 #[derive(Debug, Clone, PartialEq, Eq)]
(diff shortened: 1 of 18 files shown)


crates/query-engine/translation/src/translation/helpers.rs
```diff
@@ -90,16 +91,71 @@ pub enum CollectionInfo<'env> {
...
+    fn from(value: &'a CollectionInfo<'a>) -> Self {
+        match value {
+            CollectionInfo::Table { name, info } => FieldsInfo::Table { name, info },
+            CollectionInfo::NativeQuery { name, info } => FieldsInfo::NativeQuery { name, info },
+        }
+    }
+}
+
 impl<'request> Env<'request> {
+    pub fn with_empty<F, R>(f: F) -> R
```
soupi, line 145: This has to have some docs.

crates/query-engine/translation/src/translation/helpers.rs
```diff
@@ -116,26 +172,65 @@ impl<'request> Env<'request> {
     }
 
     /// Lookup collection or composite type.
-    pub fn lookup_composite_type(
+    pub fn lookup_fields_info(
         &self,
         type_name: &'request str,
+    ) -> Result<FieldsInfo<'request>, Error> {
+        let info =
```
soupi, line 179:
```suggestion
        // Lookup the fields of a type name in a specific order:
        // tables, then composite types, then native queries.
        let info =
```

crates/query-engine/translation/src/translation/helpers.rs
```diff
@@ -116,26 +172,65 @@ impl<'request> Env<'request> {
...
-        let it_is_a_collection = self.lookup_collection(type_name);
-
-        match it_is_a_collection {
-            Ok(collection_info) => Ok(CompositeTypeInfo::CollectionInfo(collection_info)),
-            Err(Error::CollectionNotFound(_)) => {
-                let its_a_type = self.metadata.composite_types.0.get(type_name).map(|t| {
-                    CompositeTypeInfo::CompositeTypeInfo {
-                        name: t.name.clone(),
-                        info: t.clone(),
+        let info = self
```
soupi, line 215:
```suggestion
        // Here we don't look in native queries because: ...
        let info = self
```

crates/query-engine/translation/src/translation/helpers.rs
plcplc: Ah, agreed. It's a bit non-trivial what and why we do this.

crates/query-engine/translation/src/translation/helpers.rs
plcplc: I've added some documentation to the various lookup functions that elaborate on the whats and whys of the different variations.

crates/query-engine/metadata/src/metadata/database.rs
```diff
@@ -2,32 +2,40 @@
...
-    CompositeType(String),
+    CompositeType(CompositeTypeName),
     ArrayType(Box<Type>),
 }
 
+/// Map of all known composite types.
+#[derive(Debug, Clone, PartialEq, Eq, Default)]
+pub struct CompositeTypes(pub BTreeMap<CompositeTypeName, CompositeType>);
+
+impl CompositeTypes {
```
soupi, line 25: If you are already doing `impl`s on these types, perhaps you'd prefer to impl [Default](https://doc.rust-lang.org/std/default/trait.Default.html)?

crates/query-engine/translation/src/translation/mutation/v1/delete.rs
```diff
@@ -157,7 +160,8 @@ mod tests {
         let mut arguments = BTreeMap::new();
         arguments.insert("user_id".to_string(), serde_json::Value::Number(100.into()));
 
-        let result = translate_delete(&mut state, &delete, &arguments).unwrap();
+        let result =
+            Env::with_empty(|env| translate_delete(&env, &mut state, &delete, &arguments).unwrap());
```
soupi, line 164: This really cannot be `let env = Env::new(); translate_delete(...);`?

crates/query-engine/metadata/src/metadata/database.rs
plcplc: I don't know that I like that. I don't think Default is right, since there is no way you can generally just take an empty environment and exepect to have a good time.

soupi (approved):
Looks good. Only a few nits.

crates/query-engine/metadata/src/metadata/database.rs
soupi: Ok.

crates/query-engine/translation/src/translation/mutation/v1/delete.rs
plcplc: Well, it could, if you prefer giving all the arguments that `new(..)` takes, including having a separate `Metadata` value that `new(..)` can borrow. (at least I know rustc _sometimes_ asks me to make a separate binding when I pass references to newly constructed values).

I considered this for a bit, but in the end I came up with `with_empty(..)` so the test wouldn't have to concern itself so much with what makes up an Env.

Feel free to make any changes to this to match your preferences :-)

PR #441, opened 2024-04-24
Add semicolons.

SamirTalwar:
### What

Make it clear we're returning `()`.

### How

```
$ cargo clippy --all-targets --all-features --fix
```


danieljharvey (approved):

SamirTalwar:
I find it hard to believe that adding semicolons would change behavior.

PR #444, opened 2024-04-26
Introduce configuration version 4: Support schema-qualified type

plcplc:
### What

This PR introduces the ability to refer to scalar types and complex types by a separate name in the ndc schema. This name is formed taking into account the schema the type is defined in, in the same manner as we treat tables.

The implementation is limited so far as it only includes the parsing of this format and not its construction (though doing so for testing is as simple as changing the default empty RawConfiguration value).

The reason for this is that this is already a sizable PR, so it would be good to review and integrate before it includes CLI support and upgrading of tests.
 
### How

While this sounds innocuous it actually introduces an extra level of indirection. We used to only refer to scalar types by their unqualified name in both SQL generation and NDC schema. Now we record the defined name and schema of a type and derive a new name based off these which we use to refer to the type everywhere in the generated schema.

A knock-on effect of this is that it is now more appropriate to produce a dedicated "ScalarTypes" object in introspection that includes both name data and aggregation functions and comparison operators for scalar types in nested objects. This actually simplifies the business logic of the dataconnecter itself somewhat as well, since this new structure has the same shape as the schema-response.

### DIFF
--- a/crates/configuration/src/error.rs
+++ b/crates/configuration/src/error.rs
@@ -27,4 +27,10 @@ pub enum Error {
 
     #[error("I/O error: {0}")]
     IoError(#[from] std::io::Error),
+
+    #[error("Did not find expected version tag: \"{0}\"")]
+    DidNotFindExpectedVersionTag(String),
+
+    #[error("Unable to parse any configuration versions: TODO")]
+    UnableToParseAnyVersions(Vec<Error>),
 }
--- a/static/schema.json
+++ b/static/schema.json
@@ -315,6 +315,315 @@
           ]
         }
       }
+    },
+    {
+      "description": "Initial configuration, just enough to connect to a database and elaborate a full 'Configuration'.",
+      "type": "object",
+      "required": ["connectionSettings", "version"],
+      "properties": {
+        "version": {
+          "type": "string",
+          "enum": ["4"]
+        },
+        "$schema": {
+          "description": "Jsonschema of the configuration format.",
+          "default": null,
+          "type": ["string", "null"]
+        },
+        "connectionSettings": {
+          "description": "Database connection settings.",
+          "allOf": [
+            {
+              "$ref": "#/definitions/DatabaseConnectionSettings2"
+            }
+          ]
+        },
+        "metadata": {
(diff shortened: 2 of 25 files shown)


static/schema.json
soupi, line 1: I have a question about how the schema behaves in practice. I fear it would mash version3 and version4 together and we won't get a useful editing experience. Am I mistaken? If not, perhaps we should separate the schema to two separate files, one for each version?

soupi: > I don't think this will cause any issue in an editor that just uses a plan schema-aware json LSP server.

Could you perhaps try it out so we can be sure? I guess we need to initialize a dir, update the project, and then maybe try editing the parts that are different (e.g. scalar types) and see if the experience is as expected or it gives potentially wrong suggestions.

And what do you think about the idea of having separate schema.json for each version?

plcplc: > We can also pair and try together if you'd like.

I'd love to. (setting up such a thing would be a green-field thing for me also)

> And what do you think about the idea of having separate schema.json for each version?

I can see that working out nicely, since a configuration.json is only ever at one version at a time, and the act of updating it should be assisted by the CLI plugin anyway, which can also be tasked with upgrading the schema file.

But at any rate I think it's probably best to punt this issue to when we'll be looking holistically on how the configuration format should evolve (which we have a tracked issue for doing)

crates/configuration/src/error.rs
```diff
@@ -27,4 +27,10 @@ pub enum Error {
 
     #[error("I/O error: {0}")]
     IoError(#[from] std::io::Error),
+
+    #[error("Did not find expected version tag: \"{0}\"")]
+    DidNotFindExpectedVersionTag(String),
+
+    #[error("Unable to parse any configuration versions: TODO")]
```
soupi, line 34: why todo?

soupi (approved):
I think this looks good to me.

static/schema.json
soupi: We checked this together and vscode integration is able to distinguish between version according to the set version number and provide useful information and auto complete features as we want it to.

PR #450, opened 2024-05-02
When acquiring a connection, ping the db only if it has been idle

soupi:
### What

When we acquire a database connection from the pool, there's a high chance that the connection is alive and well. We add a new pool setting that will only ping the database after a certain amount of time the connection has been idle (default 60 seconds). This settings can be turned off.

### How

Adding a new pool settings option, and essentially doing this: https://docs.rs/sqlx/latest/sqlx/postgres/type.PgPoolOptions.html#example-custom-test_before_acquire-logic


SamirTalwar (approved):
Magnificent.

SamirTalwar:
@plcplc, does this affect your work on the next major configuration version?

plcplc:
> @plcplc, does this affect your work on the next major configuration version?

Thanks for the consideration. It does a bit, but not materially :-)

crates/connectors/ndc-postgres/src/state.rs
```diff
@@ -79,7 +79,25 @@ async fn create_pool(
 ) -> Result<PgPool, InitializationError> {
     let connect_options = PgConnectOptions::from_url(connection_url)
         .map_err(InitializationError::UnableToCreatePool)?;
-    PgPoolOptions::new()
+
+    let pool_options = match pool_settings.check_connection_after_idle {
+        // The default behaviour is to always ping on acquire.
```
plcplc, line 84: Actually, the default seems to be to check after 60 seconds.

plcplc:
Just had a closer look. `PoolSettings` is kept outside the version-modules (or was never moved there more likely), so changes to those affect all supported versions. I suppose we should try to avoid that pattern in general, but for a conservative addition such as what this PR contributes we shouldn't worry too much.

crates/connectors/ndc-postgres/src/state.rs
soupi: Sorry, I'll rephrase, the default *sqlx* behaviour is to always check :)

soupi, line 84:
```suggestion
        // Unless specified, sqlx will always ping on acquire.
```

crates/connectors/ndc-postgres/src/state.rs
```diff
@@ -79,7 +79,25 @@ async fn create_pool(
 ) -> Result<PgPool, InitializationError> {
     let connect_options = PgConnectOptions::from_url(connection_url)
         .map_err(InitializationError::UnableToCreatePool)?;
-    PgPoolOptions::new()
+
+    let pool_options = match pool_settings.check_connection_after_idle {
+        // Unless specified, sqlx will always ping on acquire.
```
soupi, line 84:
```suggestion
        // Unless specified otherwise, sqlx will always ping on acquire.
```

PR #451, opened 2024-05-03
Fix otel exporter environment variable for just dev

soupi:
### What

The OTEL exporter env var used by `just dev` has changed. Instead of replacing it with the new one, I've switched to using the cli flag.

### DIFF
--- a/justfile
+++ b/justfile
@@ -59,7 +59,7 @@ dev: start-dependencies
 dev-cockroach: start-dependencies
   CONNECTION_URI='{{ COCKROACH_CONNECTION_URI }}' \
   RUST_LOG=INFO \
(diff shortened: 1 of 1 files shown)


i-am-tom (approved):

SamirTalwar:
FWIW I plan on removing this flag soonish from ndc-sdk-rs in favor of sharing tracing support with the other components.

The approach we have taken is quite custom; OpenTelemetry recommend always using environment variables for configuration. There's a few reasons for this: they're more composable than CLI arguments, they are typically easier to vary across environments, and secret stores typically have better support for them.

I think this change is fine, but I will have to revert it at some point.

soupi:
@SamirTalwar the reason I changed to cli option is that next time something changes we will know :)

PR #452, opened 2024-05-03
Add support for upgrading version 3 configurations to version 4

plcplc:
### What

This PR adds the ability to upgrade a configuration directory in version 3 to an equivalent one in version 4.

The cli now supports the command `cli upgrade --from-dir <existing configuration dir> --to-dir <new configuration dir>`, with the intended semantics that it upgrades the configuration in `<existing configuration dir>` to the newest supported version.

### How

Many places in the code (though mostly tests) is tightly coupled with internal details of version 3 and will attempt to construct and manipulate configurations directly.

This PR attempts to define a version-agnostic interface that can be used instead.


crates/cli/src/lib.rs
```diff
@@ -139,35 +128,23 @@ async fn update(context: Context<impl Environment>) -> anyhow::Result<()> {
...
+        let output =
+            configuration::introspect(existing_configuration.clone(), &context.environment).await?;
 
         // Check that the input file did not change since we started introspecting,
-        let input_again_before_write: configuration::RawConfiguration = {
-            let configuration_file_contents =
-                read_config_file_contents(&configuration_file_path).await?;
-            serde_json::from_str(&configuration_file_contents)?
-        };
+        let input_again_before_write =
```
plcplc, line 137: Previously, this function only considered changes to specifically `configuration.json`, meaning changes to e.g. native queries in files wouldn't get caught.

crates/configuration/src/version3/metadata/native_queries.rs
```diff
@@ -244,13 +244,19 @@ impl JsonSchema for NativeQueryParts {
...
+) -> Result<NativeQuerySql, String> {
+    let contents: String = match fs::read_to_string(absolute_configuration_directory.join(file)) {
         Ok(ok) => Ok(ok),
-        Err(err) => Err(format!("{}: {}", &file.display(), err)),
+        Err(err) => Err(format!("{}: {}", file.display(), err)),
     }?;
     let sql = parse_native_query(&contents);
-    Ok(NativeQuerySql::FromFile { file, sql })
+    Ok(NativeQuerySql::FromFile {
+        file: file.to_path_buf(),
```
plcplc, line 257: We need to only record the relative path to the native query sql file

crates/configuration/src/version3/mod.rs
```diff
@@ -1,9 +1,9 @@
 //! Internal Configuration and state for our connector.
 
-mod comparison;
+pub(crate) mod comparison;
```
plcplc, line 3: These need to be more visible than before so that `upgrade_from_v3` can access them.

crates/tests/databases-tests/src/postgres/configuration_tests.rs
```diff
@@ -15,7 +15,8 @@ use tests_common::common_tests;
 
 #[tokio::test]
 async fn get_configuration_schema() {
-    let schema = schemars::schema_for!(ndc_postgres_configuration::RawConfiguration);
```
plcplc, line 18: Schemas used to contain the union of all supported versions. Now they relate only to a single version (like they used to when there was only a single version)

crates/tests/tests-common/src/common_tests/configuration_v3_tests.rs
```diff
@@ -75,11 +75,7 @@ async fn read_configuration(
     let contents =
         fs::read_to_string(absolute_configuration_directory.join("configuration.json")).await?;
 
-    let mut multi_version: serde_json::Value = serde_json::from_str(&contents)?;
-
-    // We assume the stored NDC metadata file to be in the newest version, so to be able to make
```
plcplc, line 80: The check for version tags have become the responsibility of each version implementation. Therefore we shouldn't remove it.

crates/configuration/src/configuration.rs
```diff
@@ -2,33 +2,36 @@
...
+///
+/// It supports various uses:
+///
+/// * It can be turned into a `Configuration`, to be used at runtime.
+/// * It retains all information necessary to produce an equivalent serialized representation.
+/// * It supports updates between versions which may require more detailed information than is
+///   available in a `Configuration` (such as whether a native query was defined inline or in a
+///   file)
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub enum ParsedConfiguration {
```
plcplc, line 27: `RawConfiguration` is now called `ParsedConfiguration`, and it is no longer generically serializable.

Instead it defers to the implementations of different versions for ways to parse/write/maniuplate configuration value instances.

crates/configuration/src/version4/upgrade_from_v3.rs
soupi, line 1: I would love to see docstring on these new modules and their functions. I know it might be a bit of a bother, but it pays off.

The most important things to document in this particular module is "what is this module doing" and "what is actually different between v3 and v4".

crates/configuration/src/configuration.rs
```diff
@@ -47,31 +50,60 @@ pub struct Configuration {
...
+            Err(v3_err) => Err(ParseConfigurationError::UnableToParseAnyVersions(
+                MultiError(vec![Box::new(v4_err), Box::new(v3_err)]),
+            )),
+            Ok(config) => Ok(ParsedConfiguration::Version3(config)),
+        },
+        Ok(config) => Ok(ParsedConfiguration::Version4(config)),
+    }
+}
+
+pub fn make_runtime_configuration(
```
soupi, line 81: Could you please add a comment about what this do/what's the difference between a parsed conf and a runtime conf?

crates/configuration/src/version4/upgrade_from_v3.rs
plcplc: I agree - and thanks for pointing it out :-)

It is indeeed somewhat overwhelming to carry out a version bump, so these things tend to get pushed to the side sometimes.

crates/configuration/src/version4/mod.rs
```diff
@@ -405,549 +419,93 @@ fn base_type_representations() -> BTreeMap<String, database::TypeRepresentation>
...
-            .into_iter()
-            .map(|(k, uniqueness_constraint)| {
-                (k, convert_uniqueness_constraint(uniqueness_constraint))
-            })
-            .collect(),
+    // create the jsonschema file
+    let configuration_jsonschema_file_path = out_dir
+        .as_ref()
+        .to_owned()
+        .join(CONFIGURATION_JSONSCHEMA_FILENAME);
```
soupi, line 499: Wondering if each version should have a unique schema.json filename, like `schema-v3.json`, `schema-v4.json`, etc...

crates/configuration/src/version4/to_runtime_configuration.rs
```diff
@@ -0,0 +1,512 @@
...
+    // let (scalar_types, composite_types) = transitively_occurring_types(
+    //     occurring_scalar_types(
+    //         &metadata.tables,
+    //         &metadata.native_queries,
+    //         &metadata.aggregate_functions,
+    //     ),
+    //     &occurring_composite_types(&metadata.tables, &metadata.native_queries),
+    //     metadata.composite_types,
+    // );
+    //
```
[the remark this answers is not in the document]
plcplc: These are to be possibly re-introduced in a follow-up PR before we release v4, so I don't think we should remove them just yet.

crates/configuration/src/version4/mod.rs
plcplc: As long as `configuration.json` indicates which schema file to use then I suppose we could this.

I wouldn't think too hard about this until we take on actually intentionally designing the configuration format.

crates/configuration/src/configuration.rs
plcplc: Yeah, good idea. There are certainly some important nuances that it would serve us well to document. Some of this is in https://github.com/hasura/ndc-postgres/pull/452#discussion_r1589400118 but it ought to be in the code as well.

crates/configuration/src/version4/mod.rs
```diff
@@ -405,549 +419,93 @@ fn base_type_representations() -> BTreeMap<String, database::TypeRepresentation>
...
-pub fn convert_tables(tables: metadata::TablesInfo) -> query_engine_metadata::metadata::TablesInfo {
-    query_engine_metadata::metadata::TablesInfo(
-        tables
-            .0
-            .into_iter()
-            .map(|(k, table_info)| (k, convert_table_info(table_info)))
-            .collect(),
-    )
-}
+pub async fn write_parsed_configuration(
```
soupi, line 456: So we have a parsed configuration, that is one after `ndc-postgres-cli update` and writes the parsed configuration into the new directory, as well as copying the native query files from the source dir to the out dir?

soupi (approved):
Hats off for getting this cumbersome work done.
Everything seems pretty straightforward I think, two things I would appreciate:
1. Making sure that this is the interface we want with the CLI/UX team
2. More doc comments as I've mentioned already.

PR #457, opened 2024-05-07
Migrate tests to access configuration through public interface

plcplc:
### What

This PR ensures that all test code that relies on NDC configurations does so via the public interface rather than by directly depending on a specific configuration version format module.

The tests that have the version format as their subject matter have been moved into the configuration crate, cohabiting with the code they test.

The tests that don't have the version format as their subject matter now rely only on the public interface of the configuration crate. The biggest consequence of this by diff volume is that what used to be `tables.json` holding a `Metadata` value now becomes full `configuration.json` files, with their original content in the `metadata` section and a `version: "3"` tag as well.


crates/tests/databases-tests/src/cockroach/snapshots/databases_tests__cockroach__query_tests__basic__select_composite_column_simple.snap
soupi, line 1: Why were these removed? Do we not run them anymore?

soupi (approved):
Looks good to me. One comment about a test :)

crates/tests/databases-tests/src/cockroach/snapshots/databases_tests__cockroach__query_tests__basic__select_composite_column_simple.snap
plcplc: So, because I was moving tests from the database-tests crate (I think) to the configuration crate their snapshot files would move too.

What I did was simply remove every snapshot file in the repo, re-run the tests and record new snapshots. When the dust had settled these two files were gone.

And then I remembered vaguely that I had disabled some tests of composite columns in cockroach previously because it turned out they were misconfigured somehow and configuring them correctly revealed that cockroach didn't support their use of composite types after all. So I figured these were just leftovers from that.

plcplc: Ah yes:

```
    #[ignore = "Cockroach v23.2 does not introspect composite types."]
    #[tokio::test]
    async fn select_composite_column_simple() {
        let result = run_query(create_router().await, "select_composite_column_simple").await;
        insta::assert_json_snapshot!(result);
    }
```

PR #462, opened 2024-05-13
Migrate tests to configuration version 4

plcplc:
### What

This PR completes the implementation of configuration format version 4.

### How

* Deriving the set of occurring types happens in SQL now, which is remarkably easier to achieve than in rust.
* The v4 chinook deployments are also tested for stability similar to v3
* All translation tests and db tests now use the V4 configuration format. This was migrated automatically using the cli upgrade command, helping to make it more robust.
* The default type representations are now recorded in `introspectionOptions` so user preferences can be persisted across updates.


danieljharvey (approved):
We went through this together and this is great, all makes sense, and pushes a lot of the pain into the introspection SQL rather than the Rust code, which is ideal. Tremendous! 🚀

PR #463, opened 2024-05-15
bugfix: duplicate results when filtering, over-filtering on lack of rows

soupi:
### What

We are fixing two bugs:

1. Fix returning duplicate results when filtering https://github.com/hasura/graphql-engine/issues/10238
2. Fix over-filtering when a related table contains no results (due to the use of inner join)

### How

In order to filter by a nested relationship column, we first query that nesting to fetch the relevant column, and then build our filter.

Before, we fetched the tables at the top-level using an inner join, then added a where clause.

In this PR we change this generation scheme to use `WHERE EXISTS (SELECT 1 ...)` instead.
The fetching of columns will appear inside the `WHERE` clause, each path element will have it's own select (where the nesting are combined using inner joins) and separate path elements are combined with a FULL OUTER JOIN instead of inner join.

The `EXISTS` part helps us avoid duplicates - while inner joins at the top-level can create additional rows, an EXISTS cannot. It will just inform if a particular row matches the predicate.

The `FULL OUTER JOIN` part helps us avoid over-eager filtering. See the [note here](https://github.com/hasura/ndc-postgres/pull/463#discussion_r1601884534).

### DIFF
--- a/crates/query-engine/translation/src/translation/query/filtering.rs
+++ b/crates/query-engine/translation/src/translation/query/filtering.rs
@@ -25,7 +25,7 @@ pub fn translate_expression(
     root_and_current_tables: &RootAndCurrentTables,
     predicate: &models::Expression,
 ) -> Result<sql::ast::Expression, Error> {
-    // Fetch the filter expression and the relevant joins
+    // Fetch the filter expression and the relevant joins.
     let (filter_expression, joins) =
         translate_expression_with_joins(env, state, root_and_current_tables, predicate)?;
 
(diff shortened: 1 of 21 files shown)


crates/query-engine/translation/tests/snapshots/tests__select_where_album_id_equals_self_nested_object_relationship.snap
```diff
@@ -30,100 +30,116 @@ FROM
...
+                            (
+                              "%2_BOOLEXP_Album"."Title" = cast($1 as "pg_catalog"."varchar")
+                            )
+                            AND (
+                              "%0_Track"."AlbumId" = "%2_BOOLEXP_Album"."AlbumId"
+                            )
+                          )
+                      ) AS "%2_BOOLEXP_Album"
+                  ) AS "%3_BOOLEXP_Album" FULL
+                  OUTER JOIN LATERAL (
```
soupi, line 96: Say we want to write a query with a predicate like:

```graphql
where: {
  _or: [
    {Albums: {Title: {_ilike: "%e%"}}},
    {Managers: {Name: {_ilike: "A%"}}}
  ]
}
```

The SQL predicate will be `Album.Title ilike '%e%' OR Manager.Name ilike 'A%'`.

To use this predicate, we need to extract the columns from the relationship tables.

We want to extract the titles of albums, and also extract the names of managers.
We have a few cases for the artist:

1. The artist has managers and albums
2. _ has only managers
3. _ has only albums
4. _ has neither

As a join table:

| has      | managers.name | albums.title |
| -------- | ------------- | ------------ |
| both     | value       | value      |
| managers | value      | null         |
| albums   | null          | value      |
| neither  | null          | null         |

1. inner join will only fetch row (1)
2. left join will only fetch the rows (1,2)
3. full join will fetch rows (1,2,3)

In a query with an OR expression, it is important to handle all three cases, because even if we don't have tracks, we still want to return the row if the artist has a manager named `Alfred`.

danieljharvey: I would love to see this note in the code comments, or at least a link to this PR, as it's really well explained.

crates/query-engine/translation/src/translation/query/root.rs
```diff
@@ -160,8 +160,6 @@ fn translate_query_part(
 
     select.joins.extend(relationship_joins);
 
-    select.joins.extend(filter_joins);
```
danieljharvey, line 163: This is gone because the joins are happening where?

danieljharvey: Oh - I think I just found the answer: https://github.com/hasura/ndc-postgres/pull/463/files#r1602699713

justfile
```diff
@@ -4,7 +4,7 @@ set shell := ["bash", "-c"]
 # This allows us to handle quoted arguments.
 set positional-arguments
 
-HGE_V3_DIR := env_var_or_default('HGE_V3_DIRECTORY', '../v4-engine')
+HGE_V3_DIR := env_var_or_default('HGE_V3_DIRECTORY', '../v3-engine')
```
danieljharvey, line 7: Must have gotten changed in the config version upgrade. Good catch.

danieljharvey (approved):

crates/query-engine/translation/src/translation/query/filtering.rs
```diff
@@ -231,40 +267,42 @@ pub fn translate_expression(
...
+/// >     (
+/// >         SELECT *
+/// >         FROM
+/// >            <table of path[m]> AS <fresh name>
+/// >         WHERE
+/// >            <table m join condition on table m-1>
+/// >            AND <predicate of path[m]>
+/// >     ) AS <LAST-FRESH-NAME>
+/// >   ) AS <fresh name>
+/// > )
```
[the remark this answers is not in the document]
danieljharvey: (actually, no, `or` would have been behaving like `and` because of the inner join?)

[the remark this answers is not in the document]
soupi: exactly.

crates/query-engine/translation/tests/snapshots/tests__select_where_album_id_equals_self_nested_object_relationship.snap
soupi: Added a link to this comment in the code comments :)

PR #470, opened 2024-05-21
add post-insert check constraint to experimental insert mutations

soupi:
### What

https://hasurahq.atlassian.net/browse/PG-99

We currently have experimental delete and insert implementations.
The delete implementation supports permissions, the insert one does not.

We would like to support a check permission for inserts.
The check permission should verify that the values we insert match a given Boolean Expression predicate.

### How

In V2, we create a SQL query that not only returns the results as part of the CTE, but also the boolean expression. It looks similar to this:

```sql
WITH "_mra__album" AS (
  INSERT INTO
    "public"."album" ("id", "artist_id", "title")
  VALUES
    (('10') :: integer, ('1') :: integer, ('Best Album') :: text)
  RETURNING
    *, -- below this part is the constraint check
    ((("public"."album"."artist_id") = (('10') :: integer))
      OR ((("public"."album"."artist_id") IS NULL) AND ((('1') :: integer) IS NULL))
    ) :: boolean AS "check__constraint"
),
"_aca__album" AS (
  SELECT
    "id",
    "title",
    "artist_id"
  FROM
    "_mra__album"
)
SELECT
  json_build_object(
    'affected_rows', (SELECT COUNT(*) FROM "_aca__album"),
    'returning',
    (SELECT coalesce(json_agg("root"), '[]') AS "root"
      FROM
        (
          SELECT
            row_to_json(
              (SELECT "_e" FROM (SELECT "_root.base"."title" AS "title", "_root.base"."id" AS "id") AS "_e")
            ) AS "root"
          FROM
            (SELECT * FROM "_aca__album" WHERE ('true')) AS "_root.base"
        ) AS "_root"
    )
  ), -- two columns are returned, the second one contains the constraint check
  (
    SELECT
      coalesce(bool_and("check__constraint"), 'true')
    FROM
      "_mra__album"
  )
```

We want to take a similar approach:

- The returning clause will include the boolean expression predicate
- Instead of returning just a single column with the output, we return an additional column which is the result of the predicate
- In Rust code, we check that the predicate, if it is false, we cancel the transaction and return an error

To do this, we do a few things:

1. Each generated mutation translation step returns a CTE. We now make the CTE also return an additional constraint check expression, if it is not required by the mutation, we simply set it to be a constant `true`, and return the name of this column.
2. When we translate a generated mutation into a SELECT query, we build a query that returns 2 columns instead of one - one column for the results and one for the constraint check
3. In the execution step, if the result contains 2 columns, we check the second one and see if it is false. If it is, we abort the transaction.


crates/connectors/ndc-postgres/src/mutation/explain.rs
```diff
@@ -62,10 +62,13 @@ pub async fn explain(
...
-            .flat_map(|(name, sql, plan)| {
+            // we enumerate because the procedure may be invoked multiple times
+            // in a mutation request.
+            .enumerate()
+            .flat_map(|(number, (name, sql, plan))| {
                 vec![
-                    (format!("{name} SQL Mutation"), sql),
-                    (format!("{name} Execution Plan"), plan),
+                    (format!("{number} {name} SQL Mutation"), sql),
+                    (format!("{number} {name} Execution Plan"), plan),
```
soupi, line 71: This is a change from before where explain only contained the name of the procedure.

crates/query-engine/execution/src/mutation.rs
```diff
@@ -114,17 +114,24 @@ async fn execute_query(
...
+            }
+            buffer.put(results_bytes);
+
+            // if our mutation also includes a check constraint, return it.
+            if row.len() == 2 {
+                let check_constraint = row.try_get::<bool, usize>(1)?;
+                Ok(Some(check_constraint))
+            } else {
+                Ok(None)
             }
```
soupi, line 134: fetch the check constraint field.

crates/query-engine/execution/src/mutation.rs
```diff
@@ -139,7 +146,13 @@ async fn execute_query(
             server.port = database_info.server_port,
         ))
         .await?;
-    Ok(())
+
+    match check_constraint {
+        // If we have a check constraint and it failed, throw an error.
+        Some(false) => Err(Error::Query(QueryError::MutationConstraintFailed)),
```
soupi, line 152: check the check constraint field.

crates/connectors/ndc-postgres/src/schema.rs
```diff
@@ -538,6 +538,15 @@ fn experimental_insert_to_procedure(
...
     );
+    arguments.insert(
+        insert.constraint.argument_name.clone(),
+        models::ArgumentInfo {
+            argument_type: models::Type::Predicate {
+                object_type_name: insert.collection_name.clone(),
+            },
+            description: Some(insert.constraint.description.clone()),
+        },
+    );
```
soupi, line 549: add the predicate to the schema of experimental/insert.

crates/query-engine/translation/src/translation/mutation/v1/insert.rs
```diff
@@ -78,21 +78,34 @@ pub fn translate(
...
+
+    Ok((insert, check_constraint_alias))
 }
 
 /// Check that no columns are missing, and that columns cannot be inserted to
 /// are not insertred.
 fn check_columns(
     columns: &BTreeMap<String, database::ColumnInfo>,
-    inserted_columns: &[ast::ColumnName],
+    inserted_columns: &[sql::ast::ColumnName],
```
danieljharvey, line 108: Would it be possible to do this import renaming pass in a PR before this one? There is a lot going on in this PR without (absolutely sensible, granted) changes like these.

crates/query-engine/execution/src/mutation.rs
danieljharvey: Is there only ever one row returned here? Do we not do `variable` / `foreach` for mutations?

crates/query-engine/execution/src/mutation.rs
danieljharvey: This is so neat, I love it.

crates/connectors/ndc-postgres/src/schema.rs
danieljharvey: Should these be optional / nullable? Otherwise if the user doesn't want to use permissions they have to explicitly add a `const true`-esq expression to their engine metadata, which seems like a pain.

crates/query-engine/translation/src/translation/mutation/v1/insert.rs
soupi: OK

crates/query-engine/translation/src/translation/mutation/v1/insert.rs
```diff
@@ -78,21 +78,34 @@ pub fn translate(
 
     check_columns(&mutation.columns, &columns, &mutation.collection_name)?;
 
-    let insert = ast::Insert {
+    // We add an always true constraint check to unify the mutations interface.
+    let check_constraint_alias =
+        sql::helpers::make_column_alias(sql::helpers::CHECK_CONSTRAINT_FIELD.to_string());
+    let check_constraint_value = sql::helpers::true_expr();
```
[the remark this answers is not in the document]
danieljharvey: I like this.

danieljharvey (approved):
A few Qs, but this is great.

crates/connectors/ndc-postgres/src/schema.rs
soupi: I assumed the standard usecase is to apply permissions to insert always, but I guess we could? Let's not do this as part of this PR though, and do that for `delete` as well.

danieljharvey: Yeah, I'm fine with that, we'd be weakening the contract so it wouldn't be a breaking change, so no rush.

soupi: This is the experimental mutations version, we can break it as much as we like :-)

crates/query-engine/translation/src/translation/mutation/v1/delete.rs
```diff
@@ -73,7 +73,7 @@ pub fn translate_delete(
     state: &mut crate::translation::helpers::State,
     delete: &DeleteMutation,
     arguments: &BTreeMap<String, serde_json::Value>,
-) -> Result<ast::Delete, Error> {
+) -> Result<(sql::ast::Delete, sql::ast::ColumnAlias), Error> {
```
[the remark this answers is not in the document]
soupi: We could use the newly added magic field name, but we'd have to be careful to always use the same one.

```rs
/// SQL field name to be used for a check constraint.
pub const CHECK_CONSTRAINT_FIELD: &str = "%check__constraint";
```

We only use unique monotonic counters for tables so we can make sure we don't have scope problems. For columns we just use the name as is.

soupi:
@danieljharvey this definitely looks much more digestible after separating the import rename. Will take that into consideration in the future. Thanks!

PR #471, opened 2024-05-22
Enable configuration v4 by default.

plcplc:
### What

This PR makes the CLI plugin produce version-4 configuration metadata when initializing a project.

### How

Simply changing the `initial()` trait function on `ParsedConfiguration` to produce the `Version4` case instead of the `Version3` case.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -4,6 +4,9 @@
 
 ### Added
 
+- Support for qualifying scalar types by their schema. This updates the
+  metadata configuration format version number from `"3"` to `"4"`.
+
 ### Changed
 
 - When acquiring a connection, ping the db only if it has been idle.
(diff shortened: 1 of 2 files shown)


changelog.md
```diff
@@ -4,6 +4,9 @@
 
 ### Added
 
+- Support for qualifying scalar types by their schema. This updates the
+  metadata configuration format version number from `"3"` to `"4"`.
```
soupi, line 8:
```suggestion
  metadata configuration format version number from `"3"` to `"4"`.
  ([#471](https://github.com/hasura/ndc-postgres/pull/471))
```

soupi (approved):

changelog.md
plcplc: Ah, I had already made this change when I saw this. Thanks though!

PR #477, opened 2024-05-23
Insert multiple objects instead of a single object

soupi:
### What

The currently generated insert implementation only provides a mechanism to insert a single object / row via the `_object` parameter.
In this PR we replace the `_object` parameter with an `_objects` parameter that accepts a list of objects, so we can insert multiple objects / rows (bulk insert).

### How

1. Change the schema to an array of objects
2. Make the sql::ast::Insert have a vector of vector of values, instead of just a vector of values, that might be Default as well as Expression
3. match user input as array of objects
4. Get a union of the columns from all objects, and traverse each object and insert a Default when a column is missing

### DIFF
--- a/crates/query-engine/translation/tests/goldenfiles/mutations/experimental_insert/configuration.json
+++ b/crates/query-engine/translation/tests/goldenfiles/mutations/experimental_insert/configuration.json
@@ -49,5 +49,5 @@
     "compositeTypes": {},
     "nativeQueries": {}
   },
-  "mutationsVersion": null
+  "mutationsVersion": "veryExperimentalWip"
 }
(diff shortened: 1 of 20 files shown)


crates/tests/databases-tests/src/postgres/snapshots/databases_tests__postgres__mutation_tests__basic__experimental_insert_custom_dog.snap
```diff
@@ -11,13 +11,21 @@ expression: result
...
+            "id": "2",
+            "name": "Krembo",
             "birthday": "2024-01-17",
             "height_cm": "160",
             "height_inch": "62.9921259842519685",
             "adopter_name": "Strauss"
           }
         ],
-        "affected_rows": 1
+        "affected_rows": 2
```
soupi, line 28: inserting 2 objects!

crates/query-engine/translation/tests/goldenfiles/mutations/experimental_insert/configuration.json
```diff
@@ -49,5 +49,5 @@
     "compositeTypes": {},
     "nativeQueries": {}
   },
-  "mutationsVersion": null
+  "mutationsVersion": "veryExperimentalWip"
```
i-am-tom, line 52: <3

i-am-tom (approved):
Love it, all seems very straightforward

PR #478, opened 2024-05-24
Support inserting an empty object

soupi:
### What

We should support the scenario of inserting an empty object `{}`, which means use the default for all columns.

### How

To support a table that has only generated columns such as this one:

```sql
create table t(x int generated always as (1) stored);
```

We need to generate the following SQL to insert N empty objects:

```sql
insert into t select from generate_series(1,N);
```

So in case all of the objects are empty, we translate to the query plan above.


soupi:
Oof, bugs in pg 12 and 13: https://postgrespro.com/list/thread-id/2504780

crates/query-engine/translation/tests/goldenfiles/mutations/experimental_insert_empty_objects/request.json
```diff
@@ -0,0 +1,46 @@
+{
+  "operations": [
+    {
+      "type": "procedure",
+      "name": "experimental_insert_Dog",
+      "arguments": {
+        "_objects": [{}, {}, {}, {}],
```
soupi, line 7: insert 4 empty objects

crates/query-engine/translation/tests/snapshots/tests__mutations__experimental_insert_empty_objects.snap
```diff
@@ -0,0 +1,59 @@
...
+---
+BEGIN
+ISOLATION LEVEL READ COMMITTED READ WRITE;
+
+WITH "%0_generated_mutation" AS (
+  INSERT INTO
+    "public"."Dog"
+  SELECT
+  FROM
+    generate_series(1, 4) RETURNING *,
```
soupi, line 13: generate_series of 4.

crates/tests/tests-common/goldenfiles/mutations/experimental_insert_defaults.json
```diff
@@ -0,0 +1,59 @@
+{
+  "operations": [
+    {
+      "type": "procedure",
+      "name": "experimental_insert_custom_defaults",
+      "arguments": {
+        "_objects": [{}, {}, {}, {}, {}],
```
soupi, line 7: insert 5 objects

crates/tests/databases-tests/src/postgres/snapshots/databases_tests__postgres__mutation_tests__basic__experimental_insert_defaults.snap
```diff
@@ -0,0 +1,51 @@
...
+          },
+          {
+            "id": "5",
+            "name": null,
+            "birthday": "2024-01-17",
+            "height_cm": "200",
+            "height_inch": "78.7401574803149606"
+          }
+        ],
+        "affected_rows": 5
```
soupi, line 47: 5 objects were inserted

crates/query-engine/sql/src/sql/convert.rs
```diff
@@ -282,6 +293,14 @@ impl From {
                 column.to_sql(sql);
                 sql.append_syntax(")");
             }
+            From::GenerateSeries { from, to } => {
+                sql.append_syntax("generate_series");
+                sql.append_syntax("(");
+                sql.append_syntax(&from.to_string());
```
SamirTalwar, line 299: I would really like to see `.append_usize` or something; we do this a lot and it feels error-prone because changing the underlying type will not cause a compilation failure.

crates/query-engine/sql/src/sql/convert.rs
```diff
@@ -150,38 +150,53 @@ impl Insert {
...
             sql.append_syntax("(");
-            for (index, value) in object.iter().enumerate() {
-                value.to_sql(sql);
-                if index < (object.len() - 1) {
+            for (index, column_name) in columns.iter().enumerate() {
+                sql.append_identifier(&column_name.0.to_string());
+                if index < (columns.len() - 1) {
                     sql.append_syntax(", ");
                 }
             }
```
SamirTalwar, line 161: I think this is clearer if we split it into "first" and "rest". Minor suggestion, feel free to disagree.

```suggestion
            let mut columns_iter = columns.iter();
            if let Some(column_name) = columns_iter.next() {
                sql.append_identifier(&column_name.0.to_string());
            }
            for column_name in columns_iter {
                sql.append_syntax(", ");
                sql.append_identifier(&column_name.0.to_string());
            }
```

crates/query-engine/sql/src/sql/convert.rs
```diff
@@ -150,38 +150,53 @@ impl Insert {
...
 }
 
+impl InsertFrom {
+    pub fn to_sql(&self, sql: &mut SQL) {
+        match self {
+            InsertFrom::Select(select) => select.to_sql(sql),
+            InsertFrom::Values(values) => {
+                sql.append_syntax("VALUES ");
+
+                for (index, object) in values.iter().enumerate() {
```
SamirTalwar, line 181: Likewise here.

danieljharvey (approved):
This is super neat, thanks for the helpful comments.

SamirTalwar (approved):
The changes look good, I just have some minor code style comments.

crates/query-engine/sql/src/sql/convert.rs
soupi: This pattern is used throughout the module so I'd rather stay consistent. I also find it a bit clearer? Maybe it's because I'm used to it.

crates/query-engine/sql/src/sql/convert.rs
soupi: Makes sense. I'll have a look at this in a separate PR.

PR #479, opened 2024-05-27
run just generate-configuration

soupi:
### What

We haven't run `just generate-configuration` since moving to v4. This updates the configurations.


plcplc (approved):

plcplc:
We should really do this in CI too as a required check.

PR #481, opened 2024-05-27
configuration code cleanup

soupi:
### What

Remove some leftover code comments, annotate some functions, make generate-configuration idempotent.

### How

We remove the mention of the `<>` operator in favor of `!=` in configuration. That way `_neq` is always defined once, as `!=`.


plcplc (approved):
LGTM

PR #485, opened 2024-05-28
some sql phase fixes

soupi:
### What

A few small refactorings of the sql stage:

1. Introduce `sql::string::append_<number-type>` functions for literal numbers instead of using `append_syntax`.
2. Remove a couple of unnecessary to_string()s
3. Remove box from a BTreeMap
4. Rename `Int8(i32)` to `Int4(i32)`.
5. Exhaustive list for `sql::ast::Expression` in the constant folding rewrite.
6. Use a newtype instead of a struct for column alias.


SamirTalwar (approved):
This all seems very sensible to me.

PR #486, opened 2024-05-29
ndc-spec 0.1.3 with filter/order by nested fields

soupi:
### What

1. Update ndc-spec to version [0.1.3](https://github.com/hasura/ndc-spec/releases/tag/v0.1.3).
2. Implemented the new capabilities: filter and order by nested fields.

#### Nested fields selection

We add SQL syntax for accessing an expression's field. The syntax is `(<expression>).<field>`

#### Order by nested fields

We add the field selection to the sort expression in the `ORDER BY` clause so it looks like this: `ORDER BY (column).field ASC`.
Implementation wise, we thread the field path like we do for the sort direction, and wrap the expressions in a NestedFieldSelect clause.

#### Filter by nested fields

- We directly wrap columns that contain field paths when they are used inside the predicate.
- We extract the type of a nested field by recursively looking it up in the environment.

### DIFF
--- a/crates/query-engine/translation/src/translation/error.rs
+++ b/crates/query-engine/translation/src/translation/error.rs
@@ -48,11 +48,15 @@ pub enum Error {
 
 /// Capabilities we don't currently support.
 #[derive(Debug, Clone)]
-pub enum UnsupportedCapabilities {}
+pub enum UnsupportedCapabilities {
+    FieldArguments,
+}
 
 impl std::fmt::Display for UnsupportedCapabilities {
-    fn fmt(&self, _f: &mut std::fmt::Formatter) -> std::fmt::Result {
-        todo!()
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        match self {
+            UnsupportedCapabilities::FieldArguments => write!(f, "Field arguments"),
+        }
     }
 }
 
--- a/crates/query-engine/translation/src/translation/helpers.rs
+++ b/crates/query-engine/translation/src/translation/helpers.rs
@@ -554,3 +554,30 @@ impl NativeQueries {
         }
     }
 }
+
+/// A newtype wrapper around an ndc-spec type which represents accessing a nested field.
+#[derive(Debug, Clone)]
+pub struct FieldPath(pub Vec<String>);
+
+impl From<&Option<Vec<String>>> for FieldPath {
+    fn from(field_path: &Option<Vec<String>>) -> Self {
+        FieldPath(match field_path {
+            // The option has no logical function other than to avoid breaking changes.
+            None => vec![],
+            Some(vec) => vec.clone(),
+        })
+    }
+}
+
+/// Fold an expression inside of a chain of field path accessors.
+pub fn wrap_in_field_path(
+    field_path: &FieldPath,
+    expression: sql::ast::Expression,
+) -> sql::ast::Expression {
+    field_path.0.iter().fold(expression, |expression, field| {
+        sql::ast::Expression::NestedFieldSelect {
+            expression: Box::new(expression),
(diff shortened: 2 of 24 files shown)


crates/tests/databases-tests/src/postgres/snapshots/databases_tests__postgres__query_tests__predicates__filter_by_nested_field.snap
```diff
@@ -0,0 +1,29 @@
...
+            {
+              "name": "Legolas",
+              "popularity": "200"
+            },
+            {
+              "name": "Gimli",
+              "popularity": "300"
+            }
+          ],
+          "name": "Fellowship"
```
soupi, line 24: We filtered by this nested field.

crates/tests/databases-tests/src/postgres/snapshots/databases_tests__postgres__query_tests__sorting__order_by_nested_field.snap
```diff
@@ -0,0 +1,63 @@
...
+            {
+              "name": "Legolas",
+              "popularity": "200"
+            },
+            {
+              "name": "Gimli",
+              "popularity": "300"
+            }
+          ],
+          "name": "Fellowship"
```
soupi, line 58: we sorted by this field, descending.

crates/tests/tests-common/goldenfiles/select_array_composite_field.json
```diff
@@ -12,7 +12,20 @@
...
+          "order_direction": "asc",
+          "target": {
+            "type": "column",
+            "name": "characters",
+            "path": []
+          }
+        }
+      ]
+    },
+    "limit": 1
```
soupi, line 28: We added more data to the table, this was done to keep the previous output for the test.

crates/query-engine/translation/src/translation/error.rs
```diff
@@ -48,11 +48,15 @@ pub enum Error {
 
 /// Capabilities we don't currently support.
 #[derive(Debug, Clone)]
-pub enum UnsupportedCapabilities {}
+pub enum UnsupportedCapabilities {
+    FieldArguments,
```
soupi, line 52: We don't support field arguments atm.

crates/query-engine/translation/src/translation/helpers.rs
```diff
@@ -554,3 +554,30 @@ impl NativeQueries {
...
+/// A newtype wrapper around an ndc-spec type which represents accessing a nested field.
+#[derive(Debug, Clone)]
+pub struct FieldPath(pub Vec<String>);
+
+impl From<&Option<Vec<String>>> for FieldPath {
+    fn from(field_path: &Option<Vec<String>>) -> Self {
+        FieldPath(match field_path {
+            // The option has no logical function other than to avoid breaking changes.
+            None => vec![],
+            Some(vec) => vec.clone(),
```
soupi, line 567: Potentially we could use refcast, but this is probably overkill for what is probably a small amount of fairly short vectors.

crates/query-engine/translation/src/translation/query/sorting.rs
```diff
@@ -291,10 +318,23 @@ fn translate_order_by_target_group(
 /// or a select query describing how to reach the columns.
 enum ColumnsOrSelect {
     /// Columns represents target columns that are referenced from the current table.
-    Columns(Vec<(usize, models::OrderDirection, sql::ast::ColumnReference)>),
-    /// Select represents a select query which contain the requested columns.
+    Columns(
```
plcplc, line 321: Nice!

plcplc (approved):
LGTM.

Love the improved commentary.

PR #487, opened 2024-05-31
Move all dependencies to the workspace level.

SamirTalwar:
### What

Rather than specifying dependency versions in many places, we can specify them once, at the top-level workspace, and then refer to the versions in the individual crates. This will make sure we never (unintentionally) directly depend on multiple versions of the same crate.

### How

We make use of Cargo's "workspace dependencies" feature to specify the versions, and then use `workspace = true` in the individual crates.

I have also truncated the versions we depend upon to the semver-compatible range, so that upgrades within semver only need to change the Cargo lockfile.


danieljharvey (approved):

soupi (approved):
Cool stuff.

PR #488, opened 2024-06-03
Refactor: Use functions for identifier names instead of unwrapping

soupi:
### What

Small refactor: in sql::convert, instead of using unwrapping newtypes and using sql.append_identifier for table names, column names, and schema names, just implement a dedicated function (`impl` of `to_sql`) for each and use that.


SamirTalwar (approved):
This is my jam.

PR #491, opened 2024-06-03
add UPDATE ast

soupi:
### What

We will soon add an auto generated Update by key procedure. This procedure will be translate to a SQL UPDATE statement.

In this PR we introduce this AST representation of Update statements.
Of course, this might change once we actually implement the feature.

### How

Add AST definition, conversion to string, and constant folding handling.

https://www.postgresql.org/docs/current/sql-update.html


danieljharvey (approved):
I love these small self-contained changes.

PR #492, opened 2024-06-06
Unify query part handling for rows and aggregates

soupi:
### What

fixes https://hasurahq.atlassian.net/browse/PG-112

Previously, order by and limit clauses were not applied for the aggregates part of the query as the spec said it was not needed, however, that is a mistake which we now fix.

### How

To fix this, we unify the way we handle rows and aggregates.
Callers will no longer need to call `translate_rows_query` and `translate_aggregates_query` separately and then try to stitch them together, but rather call one part, `translate_query`, which will call both.

Each translate part will call the same function, `translate_query_part` which translates the common elements (where, order by, limit, etc.), but will different in the way it handles fields or aggregates.

This way, we make sure that an ndc-spec query is translated consistently between the two parts - rows and aggregates.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -11,6 +11,9 @@
 
 ### Fixed
 
+-  Unify query part handling for rows and aggregates.
+  ([#471](https://github.com/hasura/ndc-postgres/pull/492))
+
 ## [v0.7.0] - 2024-05-22
 
 ### Added
--- a/changelog.md
+++ b/changelog.md
@@ -11,7 +11,7 @@
 
 ### Fixed
 
-- Unify query part handling for rows and aggregates.
+- Rows and aggregates parts of the query should operate on the same query parameters (where, order by, limit and offset).
   ([#471](https://github.com/hasura/ndc-postgres/pull/492))
 
 ## [v0.7.0] - 2024-05-22
--- a/changelog.md
+++ b/changelog.md
@@ -11,7 +11,7 @@
 
 ### Fixed
 
--  Unify query part handling for rows and aggregates.
+- Unify query part handling for rows and aggregates.
   ([#471](https://github.com/hasura/ndc-postgres/pull/492))
 
 ## [v0.7.0] - 2024-05-22
--- a/crates/query-engine/translation/src/translation/query/mod.rs
+++ b/crates/query-engine/translation/src/translation/query/mod.rs
@@ -30,19 +30,15 @@ pub fn translate(
         None,
         variables_table_ref,
     );
-    let (current_table, from_clause) = root::make_from_clause_and_reference(
-        &query_request.collection,
-        &query_request.arguments,
-        &env,
-        &mut state,
-        None,
-    )?;
 
-    let ((inner_select, inner_select_alias), select_set) = root::translate_query(
+    let select_set = root::translate_query(
         &env,
         &mut state,
-        current_table,
-        from_clause,
+        root::MakeFrom::Collection {
+            name: query_request.collection.clone(),
+            arguments: query_request.arguments.clone(),
+        },
+        None,
         &query_request.query,
     )?;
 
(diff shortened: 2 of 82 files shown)


crates/query-engine/translation/src/translation/query/mod.rs
```diff
@@ -30,19 +30,15 @@ pub fn translate(
...
-        None,
-    )?;
 
-    let select_set = translate_query(
+    let select_set = root::translate_query(
         &env,
         &mut state,
-        &current_table,
-        &from_clause,
+        root::MakeFrom::Collection {
```
soupi: Instead of building a from clause and passing it, we build a structure describing how to build the from clause, so it can be built the same way for both `rows` and `aggregates` but produce different aliases using the state index.

crates/query-engine/translation/src/translation/query/relationships.rs
```diff
@@ -35,87 +35,18 @@ pub fn translate_joins(
...
-            let select_set = super::translate_query(
+            let select_set = root::translate_query(
                 env,
                 state,
-                &target_collection,
-                &from_clause,
+                root::MakeFrom::Collection {
+                    name: relationship.target_collection.clone(),
+                    arguments: arguments.clone(),
+                },
```
soupi: Instead of building a from clause and passing it, we build a structure describing how to build the from clause.

crates/query-engine/translation/src/translation/query/relationships.rs
```diff
@@ -35,87 +35,22 @@ pub fn translate_joins(
...
+                // We ask to inject the join predicate into the where clause.
+                &Some(root::JoinPredicate {
+                    join_with: current_table,
+                    relationship,
+                }),
                 &join_field.query,
             )?;
 
-            // add join expressions to row / aggregate selects
```
soupi, line 56: injecting the join predicate means we don't need to unpack afterwards.

crates/query-engine/translation/tests/goldenfiles/aggregate_limit_offset_order_by/request.json
soupi, line 1: Test that limit and order by has been applied to the aggregates query.

crates/query-engine/translation/tests/snapshots/tests__aggregate_limit_offset_order_by.snap
```diff
@@ -0,0 +1,52 @@
...
+                FROM
+                  "public"."Invoice" AS "%1_Invoice"
+                WHERE
+                  (
+                    "%1_Invoice"."BillingCountry" = cast($1 as "pg_catalog"."varchar")
+                  )
+                ORDER BY
+                  "%1_Invoice"."InvoiceId" ASC
+                LIMIT
+                  10 OFFSET 5
```
soupi, line 42: yep, has been applied.

crates/tests/databases-tests/src/aurora/snapshots/databases_tests__aurora__query_tests__aggregation__aggregate_count_albums_plus_field.snap
soupi, line 1: Now we apply the limit. We also changed the test a bit so it is easy to verify.

crates/tests/databases-tests/src/aurora/snapshots/databases_tests__aurora__query_tests__sorting__select_order_by_artist_name.snap
soupi, line 1: the test request did not contain order for the album, so it was ambigious. fixed.

plcplc (approved):

changelog.md
```diff
@@ -11,6 +11,9 @@
 
 ### Fixed
 
+- Unify query part handling for rows and aggregates.
```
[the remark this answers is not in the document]
soupi: Makes sense. Let me rephrase:

```suggestion
- Rows and aggregate parts of the query should operate on the same query parameters (where, order by, limit and offset).
```

PR #493, opened 2024-06-06
experimental update mutations

soupi:
### What

We want to support update point mutation. This PR introduces a new (experimental) auto-generated procedure with 4 arguments:

- the unique key
- `_set` - which columns to set, to which values
- pre check - Which rows am I allowed to edit
- post check - What output am I allowed to produce

Example schema:

```json
{
  "name": "experimental_update_Artist_by_ArtistId",
  "description": "Update any row on the 'Artist' collection using the 'ArtistId' key",
  "arguments": {
    "ArtistId": {
      "description": "The identifier of an artist",
      "type": {
        "type": "named",
        "name": "int4"
      }
    },
    "_set": {
      "type": {
        "type": "named",
        "name": "experimental_update_Artist_by_ArtistId_object"
      }
    },
    "post_check": {
      "description": "Update permission post-condition predicate over the 'Artist' collection",
      "type": {
        "type": "predicate",
        "object_type_name": "Artist"
      }
    },
    "pre_check": {
      "description": "Update permission pre-condition predicate over the 'Artist' collection",
      "type": {
        "type": "predicate",
        "object_type_name": "Artist"
      }
    }
  },
  "result_type": {
    "type": "named",
    "name": "experimental_update_Artist_by_ArtistId_response"
  }
}
```

### How

We use the tricks we already learned from delete and insert:

- We generate update by unique key mutations like we do in delete
- We generate the same pre condition as in delete
- We parse and check the _set object similarly to how we do it for insert, though simplified because we only need to read one object.
- We generate the same post condition as in insert, and return it as an additional column.

### DIFF
--- a/crates/connectors/ndc-postgres/src/schema.rs
+++ b/crates/connectors/ndc-postgres/src/schema.rs
@@ -366,6 +366,9 @@ fn mutation_to_procedure(
         mutation::generate::Mutation::Experimental(
             mutation::experimental::Mutation::InsertMutation(insert),
         ) => experimental_insert_to_procedure(name, insert, object_types, scalar_types),
+        mutation::generate::Mutation::Experimental(
+            mutation::experimental::Mutation::UpdateMutation(update),
+        ) => experimental_update_to_procedure(name, update, object_types, scalar_types),
     }
 }
 
@@ -560,6 +563,65 @@ fn experimental_insert_to_procedure(
         arguments,
         models::Type::Named {
             name: insert.collection_name.to_string(),
+        },
+        object_types,
+        scalar_types,
+    )
+}
+
+/// Given an experimental `InsertMutation`, turn it into a `ProcedureInfo` to be output in the schema.
+fn experimental_update_to_procedure(
+    name: &String,
+    update: &mutation::experimental::update::UpdateMutation,
+    object_types: &mut BTreeMap<String, models::ObjectType>,
+    scalar_types: &mut BTreeMap<String, models::ScalarType>,
+) -> models::ProcedureInfo {
+    let mutation::experimental::update::UpdateMutation::UpdateByKey(update_by_key) = update;
+
+    let mut arguments = BTreeMap::new();
+    let object_type = make_object_type(&update_by_key.columns);
+    let object_name = format!("{name}_object");
+    object_types.insert(object_name.clone(), object_type);
+
+    arguments.insert(
+        update_by_key.by_column.name.clone(),
+        models::ArgumentInfo {
+            argument_type: column_to_type(&update_by_key.by_column),
+            description: update_by_key.by_column.description.clone(),
+        },
+    );
+    arguments.insert(
+        "_set".to_string(),
+        models::ArgumentInfo {
+            argument_type: models::Type::Named { name: object_name },
+            description: None,
+        },
+    );
+    arguments.insert(
+        update_by_key.pre_constraint.argument_name.clone(),
+        models::ArgumentInfo {
+            argument_type: models::Type::Predicate {
+                object_type_name: update_by_key.collection_name.clone(),
+            },
+            description: Some(update_by_key.pre_constraint.description.clone()),
+        },
+    );
+    arguments.insert(
+        update_by_key.post_constraint.argument_name.clone(),
+        models::ArgumentInfo {
+            argument_type: models::Type::Predicate {
+                object_type_name: update_by_key.collection_name.clone(),
+            },
+            description: Some(update_by_key.post_constraint.description.clone()),
+        },
+    );
+
+    make_procedure_type(
+        name.to_string(),
+        Some(update_by_key.description.to_string()),
+        arguments,
+        models::Type::Named {
+            name: update_by_key.collection_name.to_string(),
         },
         object_types,
         scalar_types,
--- a/crates/connectors/ndc-postgres/src/schema.rs
+++ b/crates/connectors/ndc-postgres/src/schema.rs
@@ -591,28 +591,28 @@ fn experimental_update_to_procedure(
         },
     );
     arguments.insert(
-        "_set".to_string(),
+        update_by_key.set_argument_name.clone(),
         models::ArgumentInfo {
             argument_type: models::Type::Named { name: object_name },
             description: None,
         },
     );
     arguments.insert(
-        update_by_key.pre_constraint.argument_name.clone(),
+        update_by_key.pre_check.argument_name.clone(),
         models::ArgumentInfo {
             argument_type: models::Type::Predicate {
                 object_type_name: update_by_key.collection_name.clone(),
             },
-            description: Some(update_by_key.pre_constraint.description.clone()),
+            description: Some(update_by_key.pre_check.description.clone()),
         },
     );
     arguments.insert(
-        update_by_key.post_constraint.argument_name.clone(),
+        update_by_key.post_check.argument_name.clone(),
         models::ArgumentInfo {
             argument_type: models::Type::Predicate {
                 object_type_name: update_by_key.collection_name.clone(),
             },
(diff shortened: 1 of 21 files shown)


crates/query-engine/translation/src/translation/mutation/experimental/update.rs
```diff
@@ -0,0 +1,255 @@
...
+
+/// The name and description of the constraint input argument.
+#[derive(Debug, Clone)]
+pub struct Constraint {
+    pub argument_name: String,
+    pub description: String,
+}
+
+/// Generate a update for each simple unique constraint on this table.
+pub fn generate_update_by_unique(
```
soupi, line 44: Generate update by unique keys.

crates/query-engine/translation/src/translation/mutation/experimental/update.rs
```diff
@@ -0,0 +1,255 @@
...
+            });
+
+            (name, update_mutation)
+        })
+        .collect()
+}
+
+/// Given the description of an update mutation (ie, `UpdateMutation`),
+/// and the arguments, output the SQL AST.
+pub fn translate(
```
soupi, line 91: translate update procedure into SQL.

crates/tests/tests-common/goldenfiles/mutations/experimental_insert_update_custom_dog.json
```diff
@@ -0,0 +1,143 @@
...
+    },
+    {
+      "type": "procedure",
+      "name": "experimental_update_custom_dog_by_id",
+      "arguments": {
+        "id": 1,
+        "_set": {
+          "height_cm": 300,
+          "adopter_name": "Strauss"
+        },
```
soupi, line 81: Test: update the columns

crates/tests/databases-tests/src/postgres/snapshots/databases_tests__postgres__mutation_tests__basic__experimental_insert_update_custom_dog.snap
```diff
@@ -0,0 +1,48 @@
...
+    },
+    {
+      "type": "procedure",
+      "result": {
+        "returning": [
+          {
+            "id": "1",
+            "name": "Cremebo",
+            "birthday": "2020-02-02",
+            "height_cm": "300",
```
soupi, line 39: Worked!

crates/tests/databases-tests/src/postgres/snapshots/databases_tests__postgres__explain_tests__mutation__experimental_insert_update_custom_dog.snap
```diff
@@ -0,0 +1,70 @@
...
+    (
+      ("custom"."dog"."id" = 1)
+      AND (
+        "custom"."dog"."adopter_name" = cast($2 as "pg_catalog"."text")
+      )
+    ) RETURNING *,
+    (
+      "custom"."dog"."adopter_name" = cast($3 as "pg_catalog"."text")
+    ) AS "%check__constraint"
+)
```
soupi, line 21: Beautiful SQL.

crates/connectors/ndc-postgres/src/schema.rs
soupi, line 1: We add our new mutations to the schema.

crates/query-engine/sql/src/sql/ast.rs
```diff
@@ -91,19 +91,11 @@ pub struct Delete {
 pub struct Update {
     pub schema: SchemaName,
     pub table: TableName,
-    pub alias: TableAlias,
-    pub set: BTreeMap<ColumnName, UpdateExpression>,
+    pub set: BTreeMap<ColumnName, InsertExpression>,
```
plcplc, line 94: It's nice to be able to reuse parts, but we should consider calling it something other than `InsertExpression` if it's used for both inserts and updates.

crates/query-engine/translation/src/translation/mutation/check_columns.rs
```diff
@@ -0,0 +1,85 @@
+//! Before inserting or updating, check the specified columns against the db schema.
+
+use crate::translation::error::Error;
+use query_engine_metadata::metadata::database;
+use query_engine_sql::sql;
+use std::collections::BTreeMap;
+
+/// Check that no columns are missing, and that columns cannot be inserted to
+/// are not inserted.
+pub fn check_columns(
```
plcplc, line 10: The commentary of this function still only talks about inserting, but it should also apply to updating, right?

crates/tests/databases-tests/src/postgres/snapshots/databases_tests__postgres__mutation_tests__negative__experimental_insert_artist_missing_column.snap
```diff
@@ -5,6 +5,6 @@ expression: result
 {
   "message": "Invalid request",
   "details": {
-    "detail": "Unable to insert into 'Artist'. Column 'ArtistId' is missing."
+    "detail": "Unable to mutate 'Artist'. Column 'ArtistId' is missing."
```
plcplc, line 8: Wishlist: It would be nice if errors could still relate directly to what the user was doing, rather than using more abstract terms.

crates/query-engine/translation/src/translation/mutation/experimental/generate.rs
```diff
@@ -3,26 +3,36 @@
...
 use super::insert::InsertMutation;
+use super::update::{generate_update_by_unique, UpdateMutation};
 use query_engine_metadata::metadata::database;
 use std::collections::BTreeMap;
 
 #[derive(Debug, Clone)]
 pub enum Mutation {
     DeleteMutation(DeleteMutation),
     InsertMutation(InsertMutation),
+    UpdateMutation(UpdateMutation),
```
plcplc, line 14: How does this change interact with versioning?

IIUC, If I have enabled mutations in my ndc configuration (i.e., a non-null `mutationsVersion`?) I now also get update mutations generated automatically.

Should we have a `mutations-version: v3` that includes updates? 
Can a 'local' user access these update mutations?

(FWIW, I think I'm fine if the answer is simply that we consider mutations an unstable, experimental, unreleased, work in progress and users shouldn't expect to be able to try them out anywhere but on cloud.)

plcplc (approved):
Solid stuff.
Do consider the comment about interaction with versions :-)

crates/query-engine/translation/src/translation/mutation/experimental/generate.rs
soupi: If you set your mutationsVersion to `veryExperimentalWIP` (or whatever it is called), these update mutations will be generated as part of the ndc-postgres-cli `update` command.

We do not release this feature yet for general consumption. The same is true for delete and insert as well. Once we decide to do it, we will rename the whole `experimental` module to `v2`.

crates/query-engine/sql/src/sql/ast.rs
soupi: Renamed to `MutationValueExpression`. Thanks!

crates/query-engine/translation/src/translation/mutation/check_columns.rs
soupi: Yes, revised the commentary.

crates/tests/databases-tests/src/postgres/snapshots/databases_tests__postgres__mutation_tests__negative__experimental_insert_artist_missing_column.snap
soupi: I made this part of the error message parameters, so now you can pass `"insert into"` or `"update"` from the caller.

PR #494, opened 2024-06-07
Invert the update columns syntax for update procedures

soupi:
### What

Slack thread for more context: https://hasurahq.slack.com/archives/C04NS5JCD8A/p1717760957475579

We want to introduce a different syntax for column update. Instead of bucketing by operations, we bucket by column names.

#### Old API

```json
{ 
  "_set": { "name": "Al", "address": "street"},
  "_inc": { "age": 1}, 
  "_concat": { "..." }
}
```

#### New API

```json
{ 
  "update_columns": { 
    "name": { "_set": "Al" },
    "address": { "_set": "street" },
    "age": { "_inc": 1 }
  }
}
```

This syntax is more consistent with other operations in other places, and makes it easy to augment the existing update procedure with new operations without adding arguments.

### How

Warning: this code was written at night, extra care is advised.

1. We change the name of the `_set` argument in the schema to `update_columns`
2. We introduce a new object type for each column update with the structure `{ _set: value }`
3. The type for `update_columns` is all the columns in the table mapped to their object type.
4. We parse the `update_columns` object by parsing each column and its operation, and generate a `MutationValueExpression`. 

## Versioning and changelog

No need to worry about versioning. This is all in the `veryExperimentalWIP` mutationVersion.

### DIFF
--- a/crates/tests/tests-common/goldenfiles/mutations/experimental_insert_update_custom_dog.json
+++ b/crates/tests/tests-common/goldenfiles/mutations/experimental_insert_update_custom_dog.json
@@ -75,9 +75,9 @@
       "name": "experimental_update_custom_dog_by_id",
       "arguments": {
         "id": 1,
-        "_set": {
-          "height_cm": 300,
-          "adopter_name": "Strauss"
+        "update_columns": {
+          "height_cm": { "_set": 300 },
+          "adopter_name": { "_set": "Strauss" }
         },
         "pre_check": {
           "type": "binary_comparison_operator",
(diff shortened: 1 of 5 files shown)


crates/tests/tests-common/goldenfiles/mutations/experimental_insert_update_custom_dog.json
soupi, line 1: What we want to achieve.

crates/tests/tests-common/goldenfiles/mutations/experimental_insert_update_custom_dog.json
```diff
@@ -75,9 +75,9 @@
       "name": "experimental_update_custom_dog_by_id",
       "arguments": {
         "id": 1,
-        "_set": {
-          "height_cm": 300,
-          "adopter_name": "Strauss"
+        "update_columns": {
+          "height_cm": { "_set": 300 },
+          "adopter_name": { "_set": "Strauss" }
```
soupi, line 80: What we wanted to achieve.

SamirTalwar (approved):
This looks… fine?

I won't pretend to understand it 100%, but it seems totally reasonable.

PR #500, opened 2024-06-17
generate mutations for multi-column unique constraints

soupi:
### What

A unique constraint on a table could be defined using multiple column. For example in `PlaylistTrack` the pair `(PlaylistId, TrackId)` is a unique constraint, and we will generate a procedure with the following structure:

```js
experimental_delete_PlaylistTrack_by_PlaylistId_and_TrackId(
    PlaylistId: int,
    TrackId: int,
    filter: boolexpr
)
```

and a similar one for updates.

### How

- We change the mutation type to include multiple keys instead of a single one.
- Instead of using `get_non_compound_uniqueness_constraints()`, we just traverse the constraints list a generate add an argument per unique constraint key.
- We generate cute names and descriptions by chaining the key names.
- We change the schema by iterating over the keys.
- We do the same for update and delete.
- We emit a warning and skip generating a procedure if something unexpected happens.


SamirTalwar:
Perhaps we could use the constraint name instead of chaining the column names?

soupi:
@SamirTalwar right, I did do this at first, but the constraint names are not very pretty tbh. We'll generating something like:

```diff
-      "name": "experimental_delete_Track_by_TrackId",
+      "name": "experimental_delete_Track_by_PK_Track",
```
and
```diff
-      "name": "experimental_delete_PlaylistTrack_by_PlaylistId_and_TrackId",
+      "name": "experimental_delete_PlaylistTrack_by_PK_PlaylistTrack",
```
and
```diff
-      "name": "experimental_delete_topology_layer_by_layer_id_and_topology_id",
+      "name": "experimental_delete_topology_layer_by_layer_pkey",
```

I'm not against it though if you think using the constraint name is better, but I'd like to do this as part of the next ticket which is reviewing the mutations API: https://hasurahq.atlassian.net/browse/PG-115

crates/query-engine/translation/src/translation/mutation/experimental/common.rs
```diff
@@ -0,0 +1,35 @@
+//! Some common helper functions.
+
+use std::collections::BTreeSet;
+
+/// Create a description string for keys. For example:
+/// > "'TrackId' key", or
+/// > "'TrackId', 'PlaylistId' and 'OtherId' keys".
+pub fn description_keys(keys: &BTreeSet<String>) -> String {
```
soupi, line 8: Yes, this is ugly, I know.

crates/query-engine/translation/src/translation/error.rs
```diff
@@ -201,3 +201,30 @@ impl std::fmt::Display for Error {
         }
     }
 }
+
+/// A type for translation warnings.
+#[derive(Debug, Clone)]
+pub enum Warning {
```
danieljharvey, line 207: I didn't realise you meant creating an enum for `Warning`, but this is very good and means moving to a more formal warnings API will be easier.

crates/query-engine/translation/src/translation/mutation/experimental/delete.rs
```diff
@@ -10,18 +9,21 @@ use query_engine_metadata::metadata;
...
+/// This can get us `DELETE FROM <table> WHERE column = <column_name_arg>, ...`.
 #[derive(Debug, Clone)]
 pub enum DeleteMutation {
     DeleteByKey {
         description: String,
         collection_name: String,
         schema_name: sql::ast::SchemaName,
         table_name: sql::ast::TableName,
-        by_column: metadata::database::ColumnInfo,
+        by_columns: Vec<metadata::database::ColumnInfo>,
```
danieljharvey, line 26: nit: would there be any benefit in using nonempty for this?

danieljharvey: I suppose if we're just `AND` or `OR`-ing expressions together from this, there are obvious `mempty` values, so maybe this is fine as is.

danieljharvey (approved):

crates/query-engine/translation/src/translation/error.rs
soupi: I didn't plan on it at first, but I had to use the same warning in multiple places, and it just makes it easier to have an enum in that case.

crates/query-engine/translation/src/translation/mutation/experimental/delete.rs
soupi: Actually yes. I wasn't aware rust had a nonempty vector type, and now I'm looking I don't see one in the stdlib. Not sure which one to choose?

danieljharvey: Feels pretty similar to the Haskell equivalent.

soupi: Thanks!

PR #501, opened 2024-06-19
Revise the mutations API

soupi:
### What

We want to make the api for mutations pretty. We make the following changes:
1. Rename arguments
2. prefix the argument names of unique constraint columns with `key_`

#### General design

* We generate delete, insert and update procedures for each table.
* A single insert procedure is generated per _table_ of the form:
  ```graphql
  <mutation-version>_insert_<table>(
      objects: [<object>],
      post_check: <boolexpr>
  )
  ```
  It lets us insert multiple objects and include a post check for permissions.
* A delete procedure is generated per _table X unique constraint_ of the form:
  ```graphql
  <mutation-version>_delete_<table>_by_<column_and_column_and...>(
      key_<column1>: <value>,
      key_<column2>: <value>,
      ...,
      pre_check: <boolexpr>
  )
  ```
  It lets us delete a single row using the uniqueness constraint, and contains a pre check for permissions.
* An update procedure is generated per _table X unique constraint_ of the form:
  ```graphql
  <mutation-version>_update_<table>_by_<column_and_column_and...>(
      key_<column1>: <value>,
      key_<column2>: <value>,
      ...,
      update_columns: { <column>: { _set: <value> }, ... },
      pre_check: <boolexpr>,
      post_check: <boolexpr>
  )
  ```
  It lets us update a single row using the uniqueness constraint by updating the relevant columns, and contains a pre check and post check for permissions.
  Note that the `update_columns` structure is different than the v2 version where we had `_set`, `_inc`, and other fields.
* Mutations using uniqueness constraints use the naming schema `by_column_and_column_and_column` instead of the db constraint name, because the former is far more helpful.
* If generating a mutation encounters an internal error, we skip that particular mutation instead of throwing an error so the connector can start at any situation.
* Naming collisions between the unique constraints and the update_columns / pre_check / post_check is avoided by prefixing argument names of the columns of a unique constraint with `key_`.

### DIFF
--- a/crates/connectors/ndc-postgres/src/schema.rs
+++ b/crates/connectors/ndc-postgres/src/schema.rs
@@ -426,7 +426,7 @@ fn experimental_delete_to_procedure(
     match delete {
         mutation::experimental::delete::DeleteMutation::DeleteByKey {
             by_columns,
-            filter,
+            pre_check,
             description,
             collection_name,
             table_name: _,
@@ -445,12 +445,12 @@ fn experimental_delete_to_procedure(
             }
 
             arguments.insert(
-                filter.argument_name.clone(),
+                pre_check.argument_name.clone(),
                 models::ArgumentInfo {
                     argument_type: models::Type::Predicate {
                         object_type_name: collection_name.clone(),
                     },
-                    description: Some(filter.description.clone()),
+                    description: Some(pre_check.description.clone()),
                 },
             );
 
@@ -547,7 +547,7 @@ fn experimental_insert_to_procedure(
     object_types.insert(object_name.clone(), object_type);
 
     arguments.insert(
-        "_objects".to_string(),
+        insert.objects_argument_name.clone(),
         models::ArgumentInfo {
             argument_type: models::Type::Array {
                 element_type: Box::new(models::Type::Named { name: object_name }),
@@ -556,12 +556,12 @@ fn experimental_insert_to_procedure(
         },
     );
     arguments.insert(
-        insert.constraint.argument_name.clone(),
+        insert.post_check.argument_name.clone(),
         models::ArgumentInfo {
             argument_type: models::Type::Predicate {
                 object_type_name: insert.collection_name.clone(),
             },
-            description: Some(insert.constraint.description.clone()),
+            description: Some(insert.post_check.description.clone()),
         },
     );
 
--- a/crates/connectors/ndc-postgres/src/schema.rs
+++ b/crates/connectors/ndc-postgres/src/schema.rs
@@ -437,7 +437,7 @@ fn experimental_delete_to_procedure(
 
             for column in by_columns {
                 arguments.insert(
-                    format!("{}{}", columns_prefix, column.name.clone()),
+                    format!("{}{}", columns_prefix, column.name),
                     models::ArgumentInfo {
                         argument_type: column_to_type(column),
                         description: column.description.clone(),
@@ -592,7 +592,7 @@ fn experimental_update_to_procedure(
     // by columns arguments.
     for by_column in &update_by_key.by_columns {
         arguments.insert(
-            format!("{}{}", update_by_key.columns_prefix, by_column.name.clone()),
+            format!("{}{}", update_by_key.columns_prefix, by_column.name),
             models::ArgumentInfo {
                 argument_type: column_to_type(by_column),
                 description: by_column.description.clone(),
--- a/crates/connectors/ndc-postgres/src/schema.rs
+++ b/crates/connectors/ndc-postgres/src/schema.rs
@@ -429,6 +429,7 @@ fn experimental_delete_to_procedure(
             pre_check,
             description,
             collection_name,
+            columns_prefix,
             table_name: _,
             schema_name: _,
         } => {
@@ -436,7 +437,7 @@ fn experimental_delete_to_procedure(
 
             for column in by_columns {
                 arguments.insert(
-                    column.name.clone(),
+                    format!("{}{}", columns_prefix, column.name.clone()),
                     models::ArgumentInfo {
                         argument_type: column_to_type(column),
                         description: column.description.clone(),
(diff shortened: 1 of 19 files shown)


crates/query-engine/translation/src/translation/mutation/experimental/insert.rs
```diff
@@ -20,15 +22,9 @@ pub struct InsertMutation {
     pub description: String,
     pub schema_name: sql::ast::SchemaName,
     pub table_name: sql::ast::TableName,
+    pub objects_argument_name: String,
```
soupi, line 25: so we can specify the name once instead of having magic strings.

crates/connectors/ndc-postgres/src/schema.rs
```diff
@@ -426,17 +426,18 @@ fn experimental_delete_to_procedure(
...
+            columns_prefix,
             table_name: _,
             schema_name: _,
         } => {
             let mut arguments = BTreeMap::new();
 
             for column in by_columns {
                 arguments.insert(
-                    column.name.clone(),
+                    format!("{}{}", columns_prefix, column.name.clone()),
```
SamirTalwar, line 440:
```suggestion
                    format!("{}{}", columns_prefix, column.name),
```

crates/connectors/ndc-postgres/src/schema.rs
```diff
@@ -591,7 +592,7 @@ fn experimental_update_to_procedure(
     // by columns arguments.
     for by_column in &update_by_key.by_columns {
         arguments.insert(
-            by_column.name.clone(),
+            format!("{}{}", update_by_key.columns_prefix, by_column.name.clone()),
```
SamirTalwar, line 595:
```suggestion
            format!("{}{}", update_by_key.columns_prefix, by_column.name),
```

crates/query-engine/translation/src/translation/mutation/experimental/delete.rs
```diff
@@ -119,7 +115,7 @@ pub fn translate(
                 .iter()
                 .map(|by_column| {
                     let unique_key = arguments
-                        .get(&by_column.name)
+                        .get(&format!("{}{}", columns_prefix, by_column.name))
```
SamirTalwar, line 118: Perhaps the `arguments` map should use a tuple or something as a key?

crates/query-engine/translation/src/translation/mutation/experimental/delete.rs
```diff
@@ -139,13 +135,13 @@ pub fn translate(
                 })
                 .collect::<Result<Vec<sql::ast::Expression>, Error>>()?;
 
-            // Build the `filter` argument boolean expression.
+            // Build the `pre_check` argument boolean expression.
             let predicate_json = arguments
-                .get(&filter.argument_name)
-                .ok_or(Error::ArgumentNotFound(filter.argument_name.clone()))?;
+                .get(&pre_check.argument_name)
+                .ok_or(Error::ArgumentNotFound(pre_check.argument_name.clone()))?;
```
SamirTalwar, line 141:
```suggestion
                .ok_or_else(|| Error::ArgumentNotFound(pre_check.argument_name.clone()))?;
```

crates/query-engine/translation/src/translation/mutation/experimental/insert.rs
```diff
@@ -90,14 +87,14 @@ fn translate_object_into_columns_and_values(
...
         }
-        serde_json::Value::Array(_) => Err(Error::UnexpectedStructure(
-            "array of arrays structure in insert _objects argument. Expecting an array of objects."
-                .to_string(),
-        )),
-        _ => Err(Error::UnexpectedStructure(
-            "array of values structure in insert _objects argument. Expecting an array of objects."
-                .to_string(),
-        )),
+        serde_json::Value::Array(_) => Err(Error::UnexpectedStructure(format!(
```
SamirTalwar, line 90: Why isn't this in the `Display` impl of `UnexpectedStructure`?

crates/query-engine/translation/src/translation/mutation/experimental/insert.rs
```diff
@@ -213,8 +210,10 @@ pub fn translate(
     arguments: &BTreeMap<String, serde_json::Value>,
 ) -> Result<(sql::ast::Insert, sql::ast::ColumnAlias), Error> {
     let object = arguments
-        .get("_objects")
-        .ok_or(Error::ArgumentNotFound("_objects".to_string()))?;
+        .get(&mutation.objects_argument_name)
+        .ok_or(Error::ArgumentNotFound(
```
SamirTalwar, line 214:
```suggestion
        .ok_or_else(|| Error::ArgumentNotFound(
```

soupi: What is the difference?

crates/query-engine/translation/src/translation/mutation/experimental/delete.rs
soupi: arguments coming from ndc-spec, and they are not used only for keys, but also for `update_columns`, `pre_check`, etc.

crates/query-engine/translation/src/translation/mutation/experimental/insert.rs
soupi: Your comment made sense to me. I went to have a look what would I need to parameterize UnexpectedStructure by to make this work, and the result was:

- array of arrays
- insert
- {} argument
- an array of objects

Which is essentially almost the entire message :( Might not be worth trying to parameterize in that case, and we probably don't need a dedicated error time for each unexpected structure.

crates/query-engine/translation/src/translation/mutation/experimental/insert.rs
soupi: Went to have a look. It seems like `.ok_or_else` is useful when the or_else part is an expensive computation. This isn't the case here - we need to create a closure that returns this particular data, I think this will end up doing more work and require more memory, won't it? So isn't it strictly worse than just `ok_or` in this case?

crates/query-engine/translation/src/translation/mutation/experimental/delete.rs
SamirTalwar: Got it, makes sense.

crates/query-engine/translation/src/translation/mutation/experimental/insert.rs
SamirTalwar: It's just because of the `.clone()`; I think it's probably borderline.

SamirTalwar (approved):

PR #502, opened 2024-06-19
Refactor: use NonEmpty and unify fetching of key columns

soupi:
### What

Quick refactor as a followup to https://github.com/hasura/ndc-postgres/pull/500#discussion_r1645749480:

1. We use NonEmpty key columns instead of a simple Vec and trace a warning when there are no keys.
2. We move the common code of key_columns fetching into the `common` module and use it for both delete and update.


danieljharvey (approved):
Nice!

PR #504, opened 2024-06-21
Add configuration version metrics

plcplc:
### What

This PR adds a metric representing the configuration version that a running ndc-postgres instance uses.

This will help us track adoption of new versions and inform deprecation timelines and documentation.

### How

Rather than extending the existing metrics in the query_engine crate we make a new Metrics struct in the ndc_postgres_configuration crate. This is to prevent the query_engine metrics from accumulating unrelated cruft and reduce the complexity of the module graph.


crates/configuration/Cargo.toml
```diff
@@ -18,6 +18,7 @@ sqlx = { workspace = true, features = ["json", "postgres", "runtime-tokio-rustls
 thiserror = { workspace = true }
 tokio = { workspace = true, features = ["full"] }
 tracing = { workspace = true }
+prometheus.workspace = true
```
SamirTalwar, line 21: Can you follow the format above, and sort the entries?

```suggestion
prometheus = { workspace = true }
```

crates/configuration/src/configuration.rs
```diff
@@ -10,8 +10,8 @@ use crate::error::{
     WriteParsedConfigurationError,
 };
 use crate::values::{IsolationLevel, PoolSettings};
-use crate::version3;
 use crate::version4;
+use crate::{version3, VersionTag};
```
SamirTalwar, line 14: Please split and reformat.

SamirTalwar (approved):
I have some formatting concerns.

crates/configuration/Cargo.toml
plcplc: Sure. Didn't notice what the actual change was because I had `cargo add prometheus` do it.

crates/connectors/ndc-postgres/src/state.rs
```diff
@@ -18,14 +18,16 @@ use query_engine_execution::metrics;
 pub struct State {
     pub pool: PgPool,
     pub database_info: DatabaseInfo,
-    pub metrics: metrics::Metrics,
+    pub query_metrics: metrics::Metrics,
+    pub configuration_metrics: ndc_postgres_configuration::Metrics,
```
soupi, line 22: I would prefer a single `metrics` field with a `Metrics` type, and that type could be a struct which is composed from both types.
I would even prefer all metrics to be handled in one place, but would be fine with an abstraction :)

crates/query-engine/execution/src/metrics.rs
```diff
@@ -324,19 +319,6 @@ impl Timer {
     }
 }
 
-/// A wrapper around the internal Prometheus error type to avoid exposing more
-/// than we need.
```
soupi, line 328: Why remove this?

PR #505, opened 2024-06-24
Run all unit tests on CI.

SamirTalwar:
### What

We were only running a subset of unit tests on CI. Let's run all of them.

### How

Instead of trying to list the (growing) package names, let's just filter out the database tests.

This also updates the snapshots to include the new connection URI default value, as these tests were failing.

I had to move some tests into the databases-tests package as they weren't really unit tests.


soupi (commented):
Very nice. I assume the tests are failing because we didn't start the relevant database for the configuration tests. We could:

1. Start them in the "test-unit", or
2. Add them to "test-ndc-postgres-with-database", or
3. Create a separate job for them

I think (2) or (3) are better than (1).

SamirTalwar:
@soupi, I have moved the CLI and configuration tests causing problems into `databases-tests`.

soupi (approved):
yofi

soupi:
@SamirTalwar test unit do not seem to be running

soupi:
I've made the change.

PR #506, opened 2024-06-24
Remove redundant clones.

SamirTalwar:
### What

I noticed a few extra calls to `.clone()` while working on an unrelated refactor. I want to remove them for brevity and simplicity; I don't expect a performance improvement.

This turns on the Clippy warning `redundant_clone`, which detects unnecessary calls to `.clone()` (and `.to_string()`).

It is an unstable warning and so might reports some false positives. If we find any, we can suppress the warning there.

### How

I turned the warning on and applied fixes everywhere.

I also saw a few that the lint didn't catch, which I fixed myself.


soupi (approved):
Cool.

PR #511, opened 2024-06-26
Implement native query builder in the CLI plugin

i-am-tom:
### What

To improve the DX of native queries, we want a CLI interface with which we can point the CLI at a SQL(ish - a native query) file and generate the stanzas for V3 metadata. This PR sort of does it.

### Design

Introduce the following subcommands:

```
Commands on Native Operations

Usage: ndc-postgres-cli native-operation <COMMAND>

Commands:
  list    List the existing Native Operations
  create  Create a new Native Operation from a SQL file
  delete  Delete an existing Native Operation from the configuration
  help    Print this message or the help of the given subcommand(s)

Options:
  -h, --help  Print help
```

#### Create

```
Create a new Native Operation from a SQL file

Usage: ndc-postgres-cli native-operation create [OPTIONS] --operation-path <OPERATION_PATH> --kind <KIND>

Options:
      --operation-path <OPERATION_PATH>
          Relative path to the SQL file inside the connector configuration directory
      --kind <KIND>
          Operation kind [possible values: query, mutation]
      --override
          Override the Native Operation definition if it exists
  -h, --help
          Print help
```

#### Delete

```
Delete an existing Native Operation from the configuration

Usage: ndc-postgres-cli native-operation delete --name <NAME> --kind <KIND>

Options:
      --name <NAME>  The name of the Native Operation
      --kind <KIND>  Operation kind [possible values: query, mutation]
  -h, --help         Print help
```

#### List

```
List the existing Native Operations

Usage: ndc-postgres-cli native-operation list

Options:
  -h, --help  Print help
```

### How

The user supplies a file containing SQL(ish - a native operation) code, and specifies the operation kind, and we use the sqlx `describe` method to fetch information about the operation. Specifically, whether it compiles or not, and what is the error if not, and what are the oids of the types of the arguments and returning columns.

From this information, we construct a NativeQueryInfo and add it to the configuration, then update it to fetch information about types we might not have tracked in the configuration yet.


crates/cli/Cargo.toml
```diff
@@ -17,6 +18,7 @@ serde_json = { workspace = true }
 serde_yaml = { workspace = true }
 thiserror = { workspace = true }
 tokio = { workspace = true, features = ["full"] }
+libpq = { workspace = true }
```
SamirTalwar, line 21: This may cause issues with linking such that it will run on any machine without external deps.

Please make sure it is statically linked.

plcplc: We no longer use libpq 🎉

crates/cli/src/native_operations.rs
soupi, line 1: We add the new commands

crates/configuration/src/version4/metadata/native_queries.rs
```diff
@@ -227,6 +229,23 @@ impl From<NativeQueryParts> for String {
     }
 }
 
+impl NativeQueryParts {
```
soupi, line 232: We introduce a method to prettyprint a native query.

crates/configuration/src/version4/native_operations.rs
soupi, line 1: We infer information about a Native Operation from a Native Operation SQL string.

crates/tests/tests-common/src/common_tests/configuration_tests.rs
```diff
@@ -0,0 +1,65 @@
...
+    }
+}
+
+/// Tests that configuration generation has not changed.
+///
+/// This test does not use insta snapshots because it checks the NDC metadata file that is shared
+/// with other tests.
+///
+/// If you have changed it intentionally, run `just generate-configuration`.
+pub async fn introspection_is_idempotent(
```
soupi, line 45: we moved this here from the other tests file.

i-am-tom:
I'm not allowed to review this because GitHub thinks I wrote it, but I wanted to say that I'm a big fan and would approve it if I could

soupi (approved):
obama meme

PR #512, opened 2024-06-27
refactor: split ndc-spec schema generation

soupi:
### What

I'm about to add version 2 of the auto-generated mutations, and I'd like to split the schema module so that each mutation version is in their own module. This refactor makes it easier to do so.

No code was harmed in the process.

### How

We take `schema.rs`, take the helpers functions and put them in `schema/helpers.rs`, take each mutation version and put it in its own `schema/mutation/<version>.rs` module, take the mutation to procedure code and put it in `schema/mutation/mod.rs` and the rest put in `schema/mod.rs`.


i-am-tom (approved):
Love it

i-am-tom (approved):
I love this too

PR #513, opened 2024-06-27
Introduce v2 of auto-generated mutations

soupi:
### What

v2 of auto-generated mutations introduces point delete, insert and update point mutations that include permission arguments.

### Interface and Design

* We generate delete, insert and update procedures for each table.

* A single insert procedure is generated per table of the form:
  ```graphql
  v2_insert_<table>(
      objects: [<object>],
      post_check: <boolexpr>
  )
  ```
  It allows us to insert multiple objects and include a post check for permissions.

* A delete procedure is generated per table X unique constraint of the form:
  ```graphql
  v2_delete_<table>_by_<column_and_...>(
      key_<column1>: <value>,
      key_<column2>: <value>,
      ...,
      pre_check: <boolexpr>
  )
  ```
  It allows us to delete a single row using the uniqueness constraint, and contains a boolexpr for permissions.

* An update procedure is generated per table X unique constraint of the form:
  ```graphql
  v2_update_<table>_by_<column_and_...>(
      key_<column1>: <value>,
      key_<column2>: <value>,
      ...,
      update_columns: { <column>: { _set: <value> }, ... },
      pre_check: <boolexpr>,
      post_check: <boolexpr>
  )
  ```
  It allows us to update a single row using the uniqueness constraint by updating the relevant columns,
  and contains a pre check and post check for permissions.

* Mutations using uniqueness constraints use the naming schema `by_column_and_column_and_column` instead of the db constraint name,
  because the former is far more helpful.
* If generating a mutation encounters an internal error, we skip that particular mutation and trace a warning instead of throwing
  an error so the connector can start at any situation.
* Naming collisions between the unique constraints and the update_columns / pre_check / post_check is avoided by prefixing argument
  names of the columns of a unique constraint with `key_`.

### How

- We replace the current `experimental` version of mutations almost word for word (the only different is the names of the procedures, prefixing `v2` instead of `experimental`) to a new directory `v2`.
- We add a new mutations version: `v2`.
- We replace the `experimental` version of ndc-spec schema generation as well.
- We fix the tests to use v2 instead.
- experimental version is also removed from configuration version3 completely.


crates/configuration/src/version3/metadata/mutations.rs
```diff
@@ -8,5 +8,4 @@ use serde::{Deserialize, Serialize};
 #[serde(rename_all = "camelCase")]
 pub enum MutationsVersion {
     V1,
-    VeryExperimentalWip,
```
danieljharvey, line 11: I was about to say "won't this break for some people" then remembered that's why we named it like this.

static/postgres/broken-queries-ndc-metadata/schema.json
```diff
@@ -1381,7 +1381,7 @@
     "MutationsVersion": {
       "description": "Which version of the generated mutations will be included in the schema",
       "type": "string",
-      "enum": ["v1", "veryExperimentalWip"]
+      "enum": ["v1", "v2"]
```
danieljharvey, line 1384: Something to think about - do we have (and do we plan to maintain) tests for both mutations versions?

danieljharvey (approved):
I love it.

static/postgres/broken-queries-ndc-metadata/schema.json
soupi: We have a couple of tests, probably not enough. We should probably add more.

PR #514, opened 2024-06-28
Fix schema generation for insert and update input objects

soupi:
### What

For the input objects of mutations (`objects`, `update_columns`), columns that have defaults should be marked as nullable, and columns that are generated should not be listed (this one worked for insert but not update).

### How

1. Moved the schema gen helpers that are only relevant for mutations into the mutation directory
2. Moved the part that handles adding or not adding generated columns out side of insert handling, and use it in both insert and update
3. Pattern match on the column info and specify the type according to the rules in `What`.


crates/connectors/ndc-postgres/src/schema/helpers.rs
soupi, line 1: These functions moved to `mutation/helpers.rs` to make it clear that they are mutation specific.

crates/connectors/ndc-postgres/src/schema/mutation/helpers.rs
```diff
@@ -0,0 +1,177 @@
...
+    for (name, column) in columns {
+        // Add the column if it is not generated.
+        match column_to_insert_type(column) {
+            None => {}
+            Some(t) => {
+                fields.insert(
+                    name.clone(),
+                    models::ObjectField {
+                        r#type: t,
+                        description: column.description.clone(),
```
soupi, line 95: We also use the description now.

crates/tests/databases-tests/src/postgres/snapshots/databases_tests__postgres__schema_tests__schema_test__get_schema.snap
```diff
@@ -7465,20 +7441,6 @@ expression: result
             "name": "update_column_custom_dog_height_cm"
           }
         },
-        "height_in": {
```
soupi, line 7468: generated column, so not included in update_columns.

crates/tests/databases-tests/src/postgres/snapshots/databases_tests__postgres__schema_tests__schema_test__get_schema.snap
```diff
@@ -7465,20 +7441,6 @@ expression: result
...
-          "description": "Update the 'height_in' column in the 'custom_dog' collection.",
-          "type": {
-            "type": "named",
-            "name": "update_column_custom_dog_height_in"
-          }
-        },
-        "id": {
```
soupi, line 7475: identity column, so not included in update_columns.

crates/connectors/ndc-postgres/src/schema/mutation/v2.rs
```diff
@@ -117,25 +118,28 @@ pub fn update_to_procedure(
...
+            // Remember for the update_columns type
+            fields.insert(
+                column_name.clone(),
+                models::ObjectField {
+                    description: Some(format!(
+                        "Update the '{column_name}' column in the '{}' collection.",
+                        update_by_key.collection_name
+                    )),
+                    // We can not specify these.
+                    r#type: models::Type::Nullable {
```
soupi, line 136: We don't have to specify any update we don't want to, so wrapping all of these in nullable.

crates/tests/databases-tests/src/postgres/snapshots/databases_tests__postgres__schema_tests__schema_test__get_schema.snap
```diff
@@ -6688,15 +6669,21 @@ expression: result
...
+              "name": "update_column_Artist_ArtistId"
+            }
           }
         },
         "Name": {
           "description": "Update the 'Name' column in the 'Artist' collection.",
           "type": {
-            "type": "named",
-            "name": "update_column_Artist_Name"
+            "type": "nullable",
```
soupi, line 6682: update columns are now wrapped in nullable so they can be omitted.

SamirTalwar (approved):
Seems quite sensible.

PR #519, opened 2024-07-04
native operation cli delete command should write to file

i-am-tom:
### What

The delete command did not write the change to file, this fixes it.

### How

Write the configuration file after deleting.


soupi (approved):
Thanks!

PR #520, opened 2024-07-05
update to ndc-sdk-rs-0.2.1

soupi:
### What

We want to update ndc-sdk-rs to 0.2.1, and it introduces several changes:

1. Error messages https://github.com/hasura/ndc-sdk-rs/pull/15
2. Update ndc-spec to 0.1.5, which introduces breaking changes by adding newtype wrappers around string types

### How

1. For errors, follow https://github.com/hasura/ndc-sdk-rs/pull/15#issuecomment-2206231760
2. For newtypes, go into type hell and fix the bugs by trying to put the ndc-models newtypes in the data structures, as they are more descriptive.


crates/connectors/ndc-postgres/src/error/convert.rs
```diff
@@ -10,19 +10,22 @@ pub fn execution_error_to_query_error(
...
-            QueryError::MutationConstraintFailed => {
-                connector::QueryError::Other(Box::new(query_error))
+                connector::QueryError::new_unprocessable_content(&"Unprocessable content")
+                    .with_details(serde_json::Value::String(query_error.to_string()))
             }
+            QueryError::MutationConstraintFailed => connector::QueryError::new("Internal error")
+                .with_details(serde_json::Value::String(query_error.to_string())),
         },
-        Error::DB(_) => connector::QueryError::Other(Box::new(error)),
+        Error::DB(_) => connector::QueryError::new("Internal error")
```
BenoitRanque, line 27: You could pass in the error itself here, if you'd like to keep the underlying error around

crates/connectors/ndc-postgres/src/error/convert.rs
```diff
@@ -10,19 +10,22 @@ pub fn execution_error_to_query_error(
...
+            .with_details(serde_json::Value::String(error.to_string())),
```
BenoitRanque, line 28: Details should be an object! you probably just want to pass in the error to the `new` function, and omit adding details

crates/connectors/ndc-postgres/src/error/convert.rs
BenoitRanque: If you pass the error itself, eventually the message will be `error.to_string()`

SamirTalwar: All this speaks to a very strange API on the NDC SDK side.

I would like to revise this so that we're not passing around duplicate information.

crates/connectors/ndc-postgres/src/error/convert.rs
SamirTalwar: I'd also be fine with this, avoiding the string message entirely and just getting to the error itself.

crates/query-engine/metadata/src/metadata/native_queries.rs
```diff
@@ -10,27 +11,36 @@ use std::collections::BTreeMap;
...
-    pub mutations: NativeQueries,
+    pub mutations: NativeMutations,
 }
 
 impl NativeOperations {
     pub fn empty() -> Self {
         NativeOperations {
             queries: NativeQueries::empty(),
-            mutations: NativeQueries::empty(),
+            mutations: NativeMutations::empty(),
```
soupi, line 21: we introduce NativeMutations because `NativeQueries` are now a map from `CollectionName`, and `NativeMutations` are a map from `ProcedureName`.

crates/query-engine/translation/src/translation/helpers.rs
```diff
@@ -96,11 +95,11 @@ pub enum CollectionInfo<'env> {
 /// dedicated composite types.
 pub enum CompositeTypeInfo<'env> {
     Table {
-        name: &'env str,
+        name: models::CollectionName,
```
soupi, line 98: Trying to track the references when the types are newtypes becomes difficult, so we don't use references from them. This is probably not a big deal especially because they are SmolStr under the hood.

SamirTalwar (approved):

PR #522, opened 2024-07-05
config v5 with native operations rename

soupi:
### What

We introduce ndc-postgres configuration version "v5".

This version changes two things:

1. Renames the field `nativeQueries` to `nativeOperations`, and instead of containing native queries and native mutations together with a flag on each marking if it is mutation or not (`isProcedure`), we nest the in different objects, so we have:

   ```
   nativeOperations:
     queries:
     mutations:
   ```

2. We nest `scalarTypes` and `compositeTypes` under one field, `types`:

   ```
   types:
     scalar:
     composite:
   ```

### How

Might be useful to read as commits.

We first copy paste v4 to v5, then create types for `Types` and `NativeOperations` which nest the other information, then we fix all the relevant things: ndc-spec schema, cli handling, upgrading from older configs, conversion to runtime config, etc.

Then we fix the tests, including introducing new configs and changing the paths the the configs.

Then we get rid of the older configurations in the `static/<db>` directories, as they are not longer used for anything.


crates/configuration/src/version5/comparison.rs
soupi, line 1: copy paste from v4.

crates/configuration/src/version5/connection_settings.rs
soupi, line 1: copy paste from v4.

crates/configuration/src/version5/introspection.sql
soupi, line 1: copy paste from v4.

crates/configuration/src/version5/metadata/database.rs
soupi, line 1: copy paste from v4.

crates/configuration/src/version5/metadata/mod.rs
```diff
@@ -0,0 +1,24 @@
...
+/// Metadata information.
+#[derive(Clone, PartialEq, Eq, Debug, Default, Serialize, Deserialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct Metadata {
+    #[serde(default)]
+    pub tables: TablesInfo,
+    #[serde(default)]
+    pub types: Types,
+    #[serde(default)]
+    pub native_operations: NativeOperations,
```
soupi, line 23: these two are different.

crates/configuration/src/version5/metadata/mutations.rs
soupi, line 1: copy paste from v4.

crates/configuration/src/version5/metadata/native_operations.rs
```diff
@@ -0,0 +1,376 @@
...
+
+/// Metadata information of Native Operations.
+#[derive(Debug, Clone, PartialEq, Eq, Default, Serialize, Deserialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct NativeOperations {
+    /// Native Queries.
+    pub queries: NativeQueries,
+    /// Native Mutations.
+    pub mutations: NativeQueries,
+}
```
soupi, line 25: We add a wrapper around two kinds of native queries.

crates/configuration/src/version5/metadata/native_operations.rs
```diff
@@ -0,0 +1,376 @@
...
+    /// We can interpolate values using `{{variable_name}}` syntax,
+    /// such as `SELECT * FROM authors WHERE name = {{author_name}}`
+    pub sql: NativeQuerySqlEither,
+    /// Columns returned by the Native Operation
+    pub columns: BTreeMap<String, ReadOnlyColumnInfo>,
+    #[serde(default)]
+    /// Names and types of arguments that can be passed to this Native Operation
+    pub arguments: BTreeMap<String, ReadOnlyColumnInfo>,
+    #[serde(default)]
+    pub description: Option<String>,
```
soupi, line 46: There's no is procedure here anymore, since we split them to two different maps. This is generally only useful for ndc-spec schema generation to generate procedures or collections.

crates/configuration/src/version5/mod.rs
```diff
@@ -0,0 +1,300 @@
...
+    Ok(ParsedConfiguration {
+        version: Version::This,
+        schema: args.schema,
+        connection_settings: args.connection_settings,
+        metadata: metadata::Metadata {
+            tables,
+            types: metadata::Types {
+                scalar: scalar_types,
+                composite: composite_types,
+            },
```
soupi, line 189: We keep the same code we had before, just wrap them in a type before returning.

crates/configuration/src/version5/native_operations.rs
soupi, line 1: copy paste from v4.

crates/configuration/src/version5/options.rs
soupi, line 1: copy paste from v4.

crates/configuration/src/version5/to_runtime_configuration.rs
soupi, line 1: Fairly straightforward because they are almost the same shape.

crates/configuration/src/version5/upgrade_from_v4.rs
```diff
@@ -0,0 +1,617 @@
...
+        } else {
+            queries.insert(name, operation);
+        }
+    }
+
+    metadata::NativeOperations {
+        queries: metadata::NativeQueries(queries),
+        mutations: metadata::NativeQueries(mutations),
+    }
+}
```
soupi, line 160: Pretty much the only part this is interesting in the file.

crates/query-engine/translation/src/translation/helpers.rs
```diff
@@ -268,28 +282,45 @@ impl<'request> Env<'request> {
...
+            }
         }
     }
 
     /// Lookup a native query's information in the metadata.
-    pub fn lookup_native_query(
+    pub fn lookup_native_mutation(
         &self,
         procedure_name: &str,
     ) -> Result<&metadata::NativeQueryInfo, Error> {
```
soupi, line 320: only used by mutations anyway, hinted by the argument named `procedure_name`.

crates/query-engine/translation/tests/goldenfiles/mutations/v2_insert/configuration.json
soupi, line 1: We didn't switch all of the translation tests to work against version 4. Perhaps we should, but we also have coverage from the db tests.

static/schema.json
```diff
@@ -38,9 +38,14 @@
...
-        "compositeTypes": {},
-        "nativeQueries": {}
+        "types": {
+          "scalar": {},
+          "composite": {}
+        },
+        "nativeOperations": {
+          "queries": {},
+          "mutations": {}
+        }
```
soupi, line 48: The change.

crates/cli/src/native_operations.rs
```diff
@@ -162,6 +173,69 @@ async fn create(
...
+                        configuration::version5::native_operations::Kind::Mutation => configuration
+                            .metadata
+                            .native_operations
+                            .mutations
+                            .0
+                            .entry(name.clone()),
+                    } {
+                        entry.insert(new_native_operation);
+                    } else {
+                        anyhow::bail!("A Native Operation with the name '{}' already exists. To override, use the --override flag.", name);
```
SamirTalwar, line 234:
```suggestion
                        anyhow::bail!("A Native Operation with the name '{name}' already exists. To override, use the --override flag.");
```

crates/configuration/src/version5/metadata/database.rs
```diff
@@ -0,0 +1,321 @@
+//! Metadata information regarding the database and tracked information.
+
+// This code was copied from a different place that predated the introduction of clippy to the
+// project. Therefore we disregard certain clippy lints:
+#![allow(
+    clippy::enum_variant_names,
+    clippy::upper_case_acronyms,
+    clippy::wrong_self_convention
+)]
```
SamirTalwar, line 9: Feels appropriate to fix these when making breaking changes.

crates/configuration/src/version5/metadata/native_operations.rs
```diff
@@ -0,0 +1,376 @@
+//! Metadata information regarding native queries.
+
+// This code was copied from a different place that predated the introduction of clippy to the
+// project. Therefore we disregard certain clippy lints:
+#![allow(clippy::wrong_self_convention)]
```
SamirTalwar, line 5: Likewise, here.

crates/query-engine/translation/src/translation/helpers.rs
```diff
@@ -187,33 +187,47 @@ impl<'request> Env<'request> {
...
+            .get(type_name)
+            .map(|t| FieldsInfo::Table {
+                name: type_name,
+                info: t,
+            })
+            .or_else(|| {
+                self.metadata
+                    .composite_types
+                    .0
+                    .get(&metadata::CompositeTypeName(type_name.to_string()))
```
SamirTalwar, line 203: Use `::ref_cast` here, perhaps?

crates/configuration/src/configuration.rs
```diff
@@ -78,24 +80,31 @@ pub async fn introspect(
...
-            Ok(config) => Ok(ParsedConfiguration::Version3(config)),
+    match version5::parse_configuration(configuration_dir.as_ref()).await {
+        Err(v5_err) => match version4::parse_configuration(configuration_dir.as_ref()).await {
+            Err(v4_err) => match version3::parse_configuration(configuration_dir.as_ref()).await {
+                Err(v3_err) => Err(ParseConfigurationError::UnableToParseAnyVersions(
+                    MultiError(vec![
+                        ("Trying V3".to_string(), Box::new(v3_err)),
+                        ("Trying V4".to_string(), Box::new(v4_err)),
+                        ("Trying V5".to_string(), Box::new(v5_err)),
+                    ]),
```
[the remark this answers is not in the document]
soupi: I think there's a hidden assumption here that we might not have the same configuration structure in later versions.

For example, if one version calls the configuration file `configuration.json` and the other `config.json`, then we need to ask each version module, how to get the config version.

Might be worth thinking about, perhaps when @plcplc returns, but for now let's continue with this way of doing things and improve it later.

crates/configuration/src/version5/metadata/database.rs
soupi: fixed.

crates/query-engine/translation/src/translation/helpers.rs
soupi: done.

crates/cli/src/native_operations.rs
soupi: fixed.

crates/configuration/src/configuration.rs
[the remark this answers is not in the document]
SamirTalwar: That's a good point; we don't know how it will continue.

SamirTalwar (approved):

PR #527, opened 2024-07-09
Delete v3 and v4 configuration tests in favor of ndc-snapshot-metadata tests

soupi:
### What

The v3 and v4 configuration tests are a bit redundant because we have the ndc-snapshot-metadata tests. This PR removes them, introduces a snapshot for the current v4-config, and deletes the schema.json in each dir because they are also not used in favor of `static/schema.json`.


plcplc (approved):
> deletes the schema.json in each dir because they are also not used in favor of static/schema.json.

I'm not entirely sure about this though. Each configuration version has its own schema (it's not the union of all versions), so there _might_ be more to this. It also might not be used at all anymore if the only thing using them were the tests which have now been deleted.)

soupi:
@plcplc it has never been used actually.

PR #529, opened 2024-07-10
Use short SHA in multitenant PR names

danieljharvey:
### What

Small quality of life change.


paritosh-08 (approved):

SamirTalwar:
We did this because Cargo treats Git commit refs of different lengths to be different refs and therefore irreconcilable.

The easiest way to avoid this is to never use short refs.

SamirTalwar (approved):

danieljharvey:
> ~We did this because Cargo treats Git commit refs of different lengths to be different refs and therefore irreconcilable.~

Yeah, don't worry, no interest in touching the actual mechanics here.

PR #540, opened 2024-07-19
bugfix: add neq to comparison operators map

soupi:
### What

Some databases do not include `!=` in their catalog's operators list. We include `<>` as well so that `_neq` can be generated for these cases.

### How

1. Add it to the list


crates/configuration/src/version5/introspection.sql
```diff
@@ -1495,7 +1495,7 @@ WITH
     WITH
       comparison_infix_operators_mapped AS
       (
-        SELECT
+        SELECT DISTINCT ON (map.exposed_name, op.*)
```
plcplc, line 1498: You'd need to update the introspection query of v3 and v4 too if you make changes to them.

IIRC The code has to solve a similar disambiguation problem in the section that deals with implicit cast extension. There it uses LIMIT + ROW_NUMBER to emulate DISTINCT ON + ORDER BY, because cockroachdb just throws ordering of nested queries out the window (see the commentary there)

plcplc: Actually I think we should leave v3 and v4 completely untouched. We shouldn't service them at all unless for a security issue.

soupi: Removed the distinct on as we agreed that it's not necessary because each database only defines one variant of the neq operator.

static/schema.json
```diff
@@ -973,14 +970,14 @@
...
-              "$ref": "#/definitions/NativeQueries"
+              "$ref": "#/definitions/NativeMutations"
             }
           ]
         }
       }
     },
     "NativeQueries": {
-      "description": "Metadata information of Native Operations.",
+      "description": "Metadata information of Native Queries.",
```
plcplc, line 980: Is this a change we _want_ to make? (IIRC the sanctioned name is "Native Operations").

It looks like this file hasn't been updated for some time and has collected a delta.

crates/configuration/src/version3/comparison.rs
```diff
@@ -46,6 +46,11 @@ impl ComparisonOperatorMapping {
                 exposed_name: "_lt".to_string(),
                 operator_kind: OperatorKind::Custom,
             },
+            ComparisonOperatorMapping {
```
plcplc, line 49: I think we should avoid making updates to old versions of the configuration format.

In this particular case it probably doesn't matter much either way, since the operator name mapping is just an introspection option, and all we are changing is the default value. Since we stamp out the defaults when we initialize a new configuration only people who explicitly refresh their defaults or create a new configuration with an old version number will ever see these changes.

But generally we should avoid it and prefer users upgrade ther configuration instead. 

In principle this change could have to incur an update of the documentation of the configuration reference section.

plcplc (approved):

static/schema.json
soupi: > It looks like this file hasn't been updated for some time and has collected a delta.

You are right.

> Is this a change we want to make? (IIRC the sanctioned name is "Native Operations").

Yes - in config v5, we split native queries to two separate fields and nest them under native operations. This particular bit represents the `queries` field, so it makes sense to call it native queries. The `mutations` field is called native mutations as well, and they both nest under the umbrella term Native Operations:

https://github.com/hasura/ndc-postgres/blob/39d79809d77d25fc0c524f087875852784398232/static/schema.json#L956

crates/configuration/src/version3/comparison.rs
soupi: I agree with you in principle, and changing this won't make a big difference because initialize will always output v5 configs, but it would make me feel better if we do this in this particular case :)

PR #541, opened 2024-07-19
Support network_supeq and network_subeq by default

soupi:
### What

This PR lists network_supeq and network_subeq in the default allowed prefix functions, to complete support for types such as CIDR.

### How

Add them to the default list, add tests, and regenerate configs.


crates/configuration/src/version5/introspection.sql
```diff
@@ -1322,9 +1322,31 @@ WITH
...
+      'custom' AS operator_kind,
+      o.oprleft as operator_type,
+      (SELECT typname FROM pg_catalog.pg_type WHERE oid = o.oprleft LIMIT 1) as operator_type_name
+    FROM
+      pg_operator as o
+    WHERE
+      o.oprname IS NOT NULL
+      AND o.oprcode::text IS NOT NULL
+      AND o.oprleft = o.oprright
+      AND o.oprresult = (SELECT oid FROM pg_catalog.pg_type WHERE typname = 'bool' LIMIT 1)
```
soupi, line 1343: This code generates a result that looks like this:

Screenshot from 2024-07-19 16-54-44

crates/configuration/src/version5/introspection.sql
```diff
@@ -1515,8 +1537,13 @@ WITH
           USING (operator_name)
         WHERE
           op.is_infix = 't'
-		ORDER BY
-		  map.exposed_name, op.*
+          AND (
+            map.operator_type is null
+            OR map.operator_type = op.argument1_type
+            OR op.argument1_cast_chain LIKE ('% -> ' || map.operator_type_name)
```
soupi, line 1543: so using the operator name and the type, i can know that `>>=` on `inet` should be called `network_supeq` and should be an operator for the type `cidr` which has a valid cast from `cidr` to `inet`.

soupi (commented):
@plcplc I know this is quite hacky. If you have ideas on how to make this awesome, do let me know :)

crates/configuration/src/version5/introspection.sql
plcplc: Considering that this gets the `exposedName` from `pg_operator.oprcode`, which is a reference to `pg_proc.oid` (which when cast to text via `::text` yields the value `pg_proc.proname`) I should think that simply adding `network_supeq` to the list of known prefix-functions in the introspection options should make it appear (and `text_pattern_ge` and friends too)

soupi:
@plcplc changed the implementation to only add the functions to the list, as we discussed.

crates/configuration/src/version5/options.rs
```diff
@@ -155,7 +155,9 @@ fn default_introspect_prefix_function_comparison_operators() -> Vec<String> {
         "lseg_perp".to_string(),
         "network_overlap".to_string(),
         "network_sub".to_string(),
+        "network_subeq".to_string(),
         "network_sup".to_string(),
+        "network_supeq".to_string(),
         "on_pb".to_string(),
```
plcplc, line 161: :+1:

crates/tests/databases-tests/src/postgres/snapshots/databases_tests__postgres__cli_version3_tests__postgres_current_only_configure_initial_configuration_is_unchanged.snap
```diff
@@ -883,6 +883,31 @@ expression: default_configuration
         "foreignRelations": {},
         "description": null
       },
+      "custom_test_cidr": {
```
plcplc, line 886: It's a bit weird that we also update v3 and v4 test snapshots for this change, but I suppose it follows from each version using sharing the same test database definition.

plcplc (approved):

PR #545, opened 2024-07-23
Remove the aurora tests

soupi:
### What

We've come to the conclusion that this works absolutely fine and it's basically postgres, just far away, so we don't need to invest resources in testing it anymore.

### How

Remove it all.


danieljharvey (approved):
💯

PR #546, opened 2024-07-23
Remove tokio-postgres.

SamirTalwar:
### What

We just used it for parsing URLs. 

### How

We can use `Url` instead.


soupi (approved):
I'm fine with this, especially since we only use it for tests.

PR #549, opened 2024-07-29
(refactor): rename translate functions and split variables from values

soupi:
### What

Wanted to streamline the naming of the translate functions which we can do because we use qualified module names.

1. Instead of something like `sorting::translate_order_by`, just use `sorting::translate`.
2. Variables selection get their own module instead of living in `values`

### DIFF
--- a/crates/query-engine/translation/src/translation/query/fields.rs
+++ b/crates/query-engine/translation/src/translation/query/fields.rs
@@ -15,30 +15,10 @@ use crate::translation::helpers::{ColumnInfo, Env, State, TableNameAndReference}
 use query_engine_metadata::metadata::{Type, TypeRepresentation};
 use query_engine_sql::sql;
 
-/// This type collects the salient parts of joined-on subqueries that compute the result of a
-/// nested field selection.
-struct JoinNestedFieldInfo {
-    select: sql::ast::Select,
-    alias: sql::ast::TableAlias,
-}
-
-/// Translate a list of nested field joins into lateral joins.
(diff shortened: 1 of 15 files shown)


crates/query-engine/translation/src/translation/query/fields.rs
```diff
@@ -15,30 +15,10 @@ use crate::translation::helpers::{ColumnInfo, Env, State, TableNameAndReference}
...
-    joins
-        .into_iter()
-        .map(|JoinNestedFieldInfo { select, alias }| {
-            sql::ast::Join::LeftOuterJoinLateral(sql::ast::LeftOuterJoinLateral {
-                select: Box::new(select),
-                alias,
-            })
-        })
-        .collect()
-}
```
soupi, line 36: moved below the pub `translate` function.

crates/query-engine/translation/src/translation/query/values.rs
```diff
@@ -103,45 +104,16 @@ fn type_to_ast_scalar_type_name(
...
-    let projected_variable_exp = sql::ast::Expression::BinaryOperation {
-        left: Box::new(variables_reference),
-        operator: sql::ast::BinaryOperator("->".to_string()),
-        right: Box::new(sql::ast::Expression::Value(sql::ast::Value::String(
-            variable.to_string(),
-        ))),
-    };
-
-    translate_projected_variable(env, state, r#type, projected_variable_exp)
-}
```
soupi, line 130: moved to the `variables` module.

plcplc (approved):
A monotonic increment in the direction of optimal elegance.

I Love it.

PR #556, opened 2024-07-30
new-version.sh script for new version PRs

soupi:
### What

We want to automate some of the release process for ndc-postgres by introducing a new `scripts/new-version.sh` script that updates the cargo package version and updates the changelog.

Run with: `scripts/new-version.sh 1.0.2`, for example.

### How

1. A bash script to run everything
2. A sed script to change the package version
3. A Haskell script to manipulate the changelog to:
   - Change the unreleased header to the version
   - Insert a new unreleased template
   - Fix the references at the bottom of the changelog

### DIFF
--- a/scripts/changelog.hs
+++ b/scripts/changelog.hs
@@ -1,0 +1,64 @@
+-- | The purpose of script is to automate some of the actions necessarily to update the changelog
+--   before releasing a new version. It will:
+--
+--   1. Replace the `[Unreleased]` header with a `[<version>] - <date>` header
+--   2. Insert a new `[Unreleased]` template at the top
+--   3. Fix the references at the bottom
+
+{-# language LambdaCase #-}
+
+import System.Environment
+import Data.List
+
+main :: IO ()
+main = do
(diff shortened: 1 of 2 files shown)


plcplc:
As much as I enjoy using Haskell for things big and small, it does seem somewhat disproportionate to introduce a dependency on GHC just for the sake of mangling the changelog file.

Of course, doing the mangling in a shell script can be complex and error-prone too, especially if it just grows over time.

How difficult do you think it would be at present to do the mangling in shell script? At first glance it looks like a combination of sed/cat/tail could do it.

soupi:
@plcplc I would definitely accept a rewrite of this script in shell, but I don't want to do it :(

Since everyone on our team should already have ghc installed, I don't think this will be a problem in practice? Worth noting that this probably saves a couple of minutes tops, so if someone really object to having ghc installed or rewriting this script, they can do the steps manually like we've up until now.

plcplc (approved):
I don't mind it that bad :-)

I'll probably decide next time I have to do a release if I should extend `flake.nix` with ghc or rewrite the script in bash :p

PR #564, opened 2024-08-08
fix nested fields relationships

soupi:
### What

Selecting nested field relationships had a bug where it would try to look up a composite type field column in the parent collection. For example:

I have a collection `institution` with a field `staff` that is a composite type that contains the field `favorite_artist_id`. Trying to have a relationship from this field to the `Artist` table would attempt to look-up this field in `institution`, instead of in `staff`.

This PR fixes this.

### How

Before, we had a type called `TableNameAndReference` that describe the metadata name of a table and held a (relationship alias) reference to it.

- We modify this type to contain a `Source` instead of a `Name`, which can be a collection name or a nested field name. This is done so it is clearer whether we are tracking a collection or a type.
- Instead of looking up a `Collection` everywhere, we lookup /anything/ that can have fields by replacing `lookup_collection` with `lookup_fields_info`.
- When translating nested fields, instead of passing the `joins` accumulator for the parent table, we create another on the fly, and translate the joins using the table created for the nested field instead of the parent table.

We also:

- Create nicer aliases for nested fields so they are easier to track
- Add a test that is based on sql tables from the `v3-engine`, so we add a new `sql` file and update the configurations accordingly.


crates/query-engine/translation/src/translation/query/fields.rs
```diff
@@ -283,29 +283,80 @@ fn translate_nested_field(
...
+    let mut join_relationship_fields: Vec<relationships::JoinFieldInfo> = vec![];
 
-    let fields_select = translate(
+    let mut fields_select = translate(
         env,
         state,
         fields,
         &nested_field_table_reference,
         nested_field_from,
+        &mut join_relationship_fields,
```
soupi, line 347: we handle and add joins here, instead in the parent.

static/institution.sql
```diff
@@ -0,0 +1,80 @@
...
+)
+;
+-- ,(
+--   3,
+--   'University of Nowhere',
+--   null,
+--   null,
+--   ARRAY ['nothing',null],
+--   NULL
+-- );
```
soupi, line 80: commenting these out because of bugs but keeping them so it's easier to work on the bug.

https://linear.app/hasura/issue/APIPG-760/configuration-nullability-issues-in-arrays-and-composite-types

crates/query-engine/sql/src/sql/convert.rs
```diff
@@ -315,17 +315,26 @@ impl From {
...
                 sql.append_syntax("UNNEST");
                 sql.append_syntax("(");
                 expression.to_sql(sql);
                 sql.append_syntax(")");
                 sql.append_syntax(" AS ");
                 alias.to_sql(sql);
-                sql.append_syntax("(");
-                column.to_sql(sql);
-                sql.append_syntax(")");
+                if !columns.is_empty() {
```
plcplc, line 326: This change confused me a lot until I realized it is in a `From` clause.

Mulling it over a bit I think it's an alright change. But I wonder if it's ever necessary to give a column signature to `FROM UNNEST(..)`. Maybe we could do without it entirely?

crates/query-engine/translation/src/translation/query/fields.rs
plcplc: I like this change.

crates/query-engine/translation/tests/snapshots/tests__select_composite_column_complex.snap
```diff
@@ -43,20 +43,20 @@ FROM
...
                       LEFT OUTER JOIN LATERAL (
                         SELECT
                           row_to_json("%4_nested_fields") AS "collected"
                         FROM
                           (
                             SELECT
-                              "%5_nested_field_binding"."address_line_1" AS "address_line_1",
-                              "%5_nested_field_binding"."address_line_2" AS "address_line_2"
+                              "%5_make_person.result.address"."address_line_1" AS "address_line_1",
+                              "%5_make_person.result.address"."address_line_2" AS "address_line_2"
```
plcplc, line 54: Urgh, this is difficult to read.

Information-wise it's much better than the arbitrary name, but the use of `.` suggests column projection happening in-line and it's only when you look closely that you see it's actually a quoted identifier.

Maybe we should change this scheme to e.g. `->` or `[<field>]` in a follow-up PR.

crates/query-engine/translation/src/translation/query/fields.rs
```diff
@@ -283,29 +283,80 @@ fn translate_nested_field(
...
+                collection_name,
+                type_name: _,
+                field_path,
+            } => {
+                let mut field_path = field_path.0.clone();
+                field_path.push(current_column_name.clone());
+                let source = TableSource::NestedField {
+                    collection_name: collection_name.clone(),
+                    type_name: nested_field_type_name,
+                    field_path: FieldPath(field_path),
```
plcplc, line 316: It looks like this lets us build up multiple nested field lookups in one go. Are there examples of this in tests?

plcplc (approved):
Left some comments - translatign nested field selection is a complex thing.

But if it works I'm happy.

crates/query-engine/translation/src/translation/query/fields.rs
soupi: The `field_path` and `collection_name` are only used for naming the alias, so even if this isn't perfect, it wouldn't effect the query.

plcplc: I think if they're only used as a convenience we should consider making that more obvious in the code. There is a lot of stuff that has happen, so the more we can direct the reader's attention the better.

soupi: Yeah I understand. I did try to make this explicit in the type's comments:
https://github.com/hasura/ndc-postgres/pull/564/files#diff-d24352dcf20dd7c6e0b5e4f0c7dec878e49c311b976280b19008620bb936bb2eR80

But I guess there is a discoverability issue here.

crates/query-engine/sql/src/sql/convert.rs
soupi: Will have a look at it in the future. I don't remember the exact case right now, but I think we use it for variables or something where we need to specify the names.

soupi: And sorry about this, I'm working on a similar ticket and kind of got some changes mixed together.

PR #565, opened 2024-08-08
Fields of composite types are always nullable

plcplc:
### What

We used to mark fields of composite type as not-nullable in the NDC schema.

This is wrong. Nullability is a property of columns of tables, not of fields of record types.

This is also demonstrated by the following transcript:
```
postgres=# create table a_table (nullable_text text, not_nullable_text text not null);
CREATE TABLE
postgres=# create table derived_table (a_table a_table, other text);
CREATE TABLE

postgres=# insert into derived_table values (('nullable', 'not nullable'), 'other text');
INSERT 0 1
postgres=# insert into derived_table values ((null, 'not nullable'), 'other text');
INSERT 0 1

postgres=# select (a_table).* from derived_table;
 nullable_text | not_nullable_text
---------------+-------------------
 nullable      | not nullable
               | not nullable
(2 rows)

postgres=# insert into a_table select (a_table).* from derived_table;
INSERT 0 2
postgres=# select * from a_table;
 nullable_text | not_nullable_text
---------------+-------------------
 nullable      | not nullable
               | not nullable
(2 rows)

-- We can easily construct a record with (not_nullable_text=null) when on **the composite type** a_table:
postgres=# insert into derived_table values (('nullable', null), 'other text');
INSERT 0 1
postgres=# select * from derived_table;
          a_table          |   other
---------------------------+------------
 (nullable,"not nullable") | other text
 (,"not nullable")         | other text
 (nullable,)               | other text
(3 rows)

postgres=# select (a_table).* from derived_table;
 nullable_text | not_nullable_text
---------------+-------------------
 nullable      | not nullable
               | not nullable
 nullable      |
(3 rows)

-- ... But we cannot insert this into **the table** a_table.
postgres=# insert into a_table select (a_table).* from derived_table;
ERROR:  null value in column "not_nullable_text" of relation "a_table" violates not-null constraint
DETAIL:  Failing row contains (nullable, null).
```

### How

We simply make the schema endpoint always return nullable fields of composite types.


plcplc:
Strictly speaking, fixing this will require a configuration version bump, since it changes the schema :-/

soupi (approved):

PR #566, opened 2024-08-12
[Native Operations]: trim semicolon suffix

soupi:
### What

There are simple cases where we can handle native queries that end with semicolon. This might help some users. https://hasurahq.slack.com/archives/C0329ERCSA1/p1723212643398509

### How

When parsing a native query,

1. trim end
2. strip_suffix(';'), or keep the trimmed version.

### DIFF
--- a/crates/configuration/src/version5/metadata/native_operations.rs
+++ b/crates/configuration/src/version5/metadata/native_operations.rs
@@ -286,8 +286,11 @@ pub fn parse_native_query_from_file(
 }
 
 /// Parse a native query into parts where variables have the syntax `{{<variable>}}`.
-pub fn parse_native_query(string: &str) -> NativeQueryParts {
-    let vec: Vec<Vec<NativeQueryPart>> = string
+pub fn parse_native_query(string_untrimmed: &str) -> NativeQueryParts {
+    let string_trimmed = string_untrimmed.trim_end();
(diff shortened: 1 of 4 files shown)


crates/configuration/src/version5/metadata/native_operations.rs
```diff
@@ -286,8 +286,11 @@ pub fn parse_native_query_from_file(
...
 /// Parse a native query into parts where variables have the syntax `{{<variable>}}`.
-pub fn parse_native_query(string: &str) -> NativeQueryParts {
-    let vec: Vec<Vec<NativeQueryPart>> = string
+pub fn parse_native_query(string_untrimmed: &str) -> NativeQueryParts {
+    let string_trimmed = string_untrimmed.trim_end();
+    let vec: Vec<Vec<NativeQueryPart>> = string_trimmed
+        .strip_suffix(';')
+        .unwrap_or(string_trimmed)
         .split("{{")
         .map(|part| match part.split_once("}}") {
```
SamirTalwar, line 295: Any reason not to use `.flat_map` here?

SamirTalwar (approved):

crates/configuration/src/version5/metadata/native_operations.rs
soupi: Not really. I don't mind doing that as part of this PR if you wish.

PR #571, opened 2024-08-13
Array elements are always nullable

plcplc:
### What

PR #565 fixed field nullability of composite type fields, but neglected to treat the similar issue that exists for arrays.

In Postgres, any array type (e.g. `int4[]`, `text[]`, etc.) may contain elements which are null. The type system does not support declaring arrays with non-null elements.

This PR fixes that issue.

### How

Whenever we have to translate an ndc-postgres array type to an ndc-spec array type in the schema generation we annotate the element type as being nullable.

### DIFF
--- a/static/institution.sql
+++ b/static/institution.sql
@@ -66,7 +66,7 @@ ROW(
 ARRAY ['Architecture and Civil Engineering','Computer Science and Engineering','Electrical Engineering','Physics','Industrial and Materials Science'],
 ROW(
     3421,
-    3421 -- NULL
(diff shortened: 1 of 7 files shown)


static/institution.sql
```diff
@@ -69,12 +69,11 @@ ROW(
     3421 -- NULL
```
soupi, line 69: Could you also replace this?

```suggestion
    NULL
```

soupi (approved):
Cool.

PR #573, opened 2024-08-13
implemented filter by nested collection

soupi:
### What

Implementing an upcoming feature in ndc-spec:
https://github.com/hasura/ndc-spec/pull/166

We want to support filtering by a field inside an array column.

For example: I want to get all institutions that where there exists a person who's last name is "Hughes" among their staff.

This requires an update to ndc-spec and ndc-sdk-rs.

### How

When we get a request with an exists over a nested field collection, we construct a nested field selection expression, unnest the results, and use that as our new "current table".


crates/query-engine/translation/src/translation/query/fields.rs
soupi, line 1: This part is not related to this change, but stems from introducing `collection_name_and_field_path()` extractor to simplify the pattern match and reduce duplication.

danieljharvey (approved):
I think I follow what is going on here. Thank you for including clear comments throughout, it helps a lot.

Cargo.toml
```diff
@@ -31,9 +31,9 @@ similar_names = "allow"
 too_many_lines = "allow"
 
 [workspace.dependencies]
-ndc-models = { git = "https://github.com/hasura/ndc-spec.git", tag = "v0.1.5" }
-ndc-sdk = { git = "https://github.com/hasura/ndc-sdk-rs.git", tag = "v0.3.0" }
-ndc-test = { git = "https://github.com/hasura/ndc-spec.git", tag = "v0.1.5" }
+ndc-models = { git = "http://github.com/hasura/ndc-spec.git", tag = "v0.1.6" }
+ndc-sdk = { git = "https://github.com/hasura/ndc-sdk-rs.git", tag = "v0.4.0" }
+ndc-test = { git = "http://github.com/hasura/ndc-spec.git", tag = "v0.1.6" }
```
soupi, line 36:
```suggestion
ndc-models = { git = "https://github.com/hasura/ndc-spec.git", tag = "v0.1.6" }
ndc-sdk = { git = "https://github.com/hasura/ndc-sdk-rs.git", tag = "v0.4.0" }
ndc-test = { git = "https://github.com/hasura/ndc-spec.git", tag = "v0.1.6" }
```

PR #574, opened 2024-08-16
Support setting ssl client certificate information via environment variables

soupi:
### What

We'd like to support setting ssl certificate information via environment variables.

### How

1. We create a new function, `get_connect_options`, which will read both the uri and the ssl information, and use it and `connect_with` everywhere instead of `connect` with just the `uri`. We make sure all operations using configuration version 5 including the connector and the cli use `get_connect_options` (except tests).
2. We read the `client_cert`, `client_key` and `root_cert` from the environment and put them directly into the sqlx connection options. 

#### How we tested this

We used [this article](https://dev.to/danvixent/how-to-setup-postgresql-with-ssl-inside-a-docker-container-5f3) as a guide on how to set up postgres+certs with docker.

After running all of the commands, we had to do the following as well:

```sh
$ certstrap request-cert --common-name postgresdb  --domain localhost
$ cp certs/out/myCA.crt out/
$ cp certs/out/myCA.key out/
$ certstrap sign postgresdb --CA myCA
```

Then, we added the following environment variables:

```sh
$ export CLIENT_CERT="$(cat /tmp/ssl/out/postgresdb.crt)"
$ export CLIENT_KEY="$(cat /tmp/ssl/out/postgresdb.key)"
$ export ROOT_CERT="$(cat /tmp/ssl/certs/out/myCA.crt)"
```

Initialized and updated the connector:

```sh
$ mdkir /tmp/ssltest
$ CONNECTION_URI="postgresql://postgres:postgres@localhost:64009/postgres?sslmode=verify-ca" target/debug/ndc-postgres-cli --context /tmp/ssltest initialize
$ CONNECTION_URI="postgresql://postgres:postgres@localhost:64009/postgres?sslmode=verify-ca" target/debug/ndc-postgres-cli --context /tmp/ssltest update
```

Added a native query:

```sh
$ echo "select 'gil' as "name", 35 as 'age'" > /tmp/ssltest/a.sql

$ CONNECTION_URI="postgresql://postgres:postgres@localhost:64009/postgres?sslmode=verify-ca" target/debug/ndc-postgres-cli --context /tmp/ssltest native-operation create --kind query --operation-path a.sql
```

Started the connector:

```sh
CONNECTION_URI="postgresql://postgres:postgres@localhost:64009/postgres?sslmode=verify-ca" target/debug/ndc-postgres serve --configuration /tmp/ssltest
```

And ran a query:

```sh
curl -X POST \
    -H 'Host: example.hasura.app' \
    -H 'Content-Type: application/json' \
    -H 'x-hasura-role: admin' \
    http://localhost:8080/query \
    -d '{ "collection": "a", "query": { "fields": { "name": { "type": "column", "column": "name" } } }, "arguments": {}, "collection_relationships": {} }' | jq
```

### DIFF
--- a/crates/configuration/src/connect.rs
+++ b/crates/configuration/src/connect.rs
@@ -1,0 +1,66 @@
+//! Connection settings.
+
+use std::borrow::Cow;
+
+use sqlx::postgres::PgConnectOptions;
+use sqlx::ConnectOptions;
+
+use crate::environment::{Environment, Variable};
+use crate::values::{ConnectionUri, Secret};
+
+/// Get the connect options from the connection string and environment.
+pub fn get_connect_options(
+    connection_uri: &ConnectionUri,
+    environment: impl Environment,
+) -> anyhow::Result<PgConnectOptions> {
+    let uri = match &connection_uri {
+        ConnectionUri(Secret::Plain(value)) => Cow::Borrowed(value),
+        ConnectionUri(Secret::FromEnvironment { variable }) => {
+            Cow::Owned(environment.read(variable)?)
+        }
+    };
+
+    let connect_options = PgConnectOptions::from_url(&uri.parse()?)?;
+
+    let ssl = read_ssl_info(environment);
+
+    // Add ssl info if present.
+    Ok(match ssl {
+        None => connect_options,
+        Some(secret) => connect_options
+            .ssl_client_cert_from_pem(secret.certificate)
+            .ssl_client_key_from_pem(secret.key)
+            .ssl_root_cert_from_pem(secret.root_certificate),
+    })
+}
+
+/// SSL client certificate information.
+struct SslClientInfo {
+    certificate: String,
+    key: String,
+    root_certificate: Vec<u8>,
+}
+
+/// Read ssl certificate and key from the environment.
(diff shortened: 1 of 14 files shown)


crates/configuration/src/connect.rs
```diff
@@ -0,0 +1,66 @@
...
+        (Some(certificate), Some(key), Some(root_certificate))
+            if !certificate.is_empty() && !key.is_empty() && !root_certificate.is_empty() =>
+        {
+            Some(SslClientInfo {
+                certificate,
+                key,
+                root_certificate,
+            })
+        }
+        _ => None,
```
SamirTalwar, line 64: Can we log a warning if some, but not all, are `None`?

SamirTalwar (approved):
Looks great!

I would love to see an automated test at some point.

danieljharvey (approved):
Nice!

crates/configuration/src/connect.rs
soupi: Will create a ticket.

soupi: https://linear.app/hasura/issue/APIPG-774/when-some-but-not-all-ssl-client-cert-variables-are-set-emit-a-warning

PR #578, opened 2024-08-19
Support setting ssl client certificate information and ssl root certificate independently

soupi:
### What

We want to be able to set the ssl root certificate without setting the ssl client certificate information. We want to be able to set the two independently.

### How

- Split the SslClientInfo to two Optionals and rename to SslInfo. Now they can be set independently.

### DIFF
--- a/crates/configuration/src/connect.rs
+++ b/crates/configuration/src/connect.rs
@@ -68,7 +68,7 @@ fn read_ssl_info(environment: impl Environment) -> SslInfo {
             None
         }
         (_, Some(key)) if !key.is_empty() => {
-            tracing::warn!("SSL client key set without key. Ignoring.");
+            tracing::warn!("SSL client key set without certificate. Ignoring.");
             None
         }
         _ => None,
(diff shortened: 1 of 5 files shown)


crates/configuration/src/connect.rs
```diff
@@ -51,16 +59,28 @@ fn read_ssl_info(environment: impl Environment) -> Option<SslClientInfo> {
...
+    let client = match (certificate, key) {
+        (Some(certificate), Some(key)) if !certificate.is_empty() && !key.is_empty() => {
+            Some(SslClientInfo { certificate, key })
+        }
+        (Some(certificate), None) if !certificate.is_empty() => {
+            tracing::warn!("SSL client certificate set without key. Ignoring.");
+            None
+        }
+        (_, Some(key)) if !key.is_empty() => {
+            tracing::warn!("SSL client key set without key. Ignoring.");
```
SamirTalwar, line 71:
```suggestion
            tracing::warn!("SSL client key set without certificate. Ignoring.");
```

SamirTalwar (approved):

PR #588, opened 2024-08-20
Specify major ndc-postgres version to cargo update

danieljharvey:
### What

`cargo update` quite helpfully errors when we do something ambiguous like `cargo update ndc-postgres` when there's two of them, so let's disambiguate the one we want.

https://github.com/hasura/ndc-postgres/actions/runs/10468915177/job/28990809823#step:5:92

### How

Change job to `cargo update ndc-postgres@1`, essentially.


soupi (approved):
Thank you Dr. Horse.

PR #635, opened 2024-10-25
Make mutation pre/post checks optional

codedmart:
### What
- [ENG-1205](https://linear.app/hasura/issue/ENG-1205/make-mutation-precheck-and-postcheck-arguments-optional) - 60fc380dd5fde6cf8bf4d85cdb81f4ae36abc46a - Change mutation pre/post checks to be optional

PR #636, opened 2024-10-25
Make `v2_` mutation prefix configurable

danieljharvey:
### What

Pulled from https://github.com/hasura/ndc-postgres/pull/635 into own PR.

We add `mutations_prefix: Option<String>` to v5 configuration. When left `null` (for existing projects), we default to `v2_mutation_name`, but it can be replaced with `horse_mutation_name` etc.

Most people will just want to use it to remove the prefix though, and that is the default for new projects. `mutationsPrefix: ""` will get rid of the prefix altogether, which I assume most users would want.


static/postgres/v5-configuration/configuration.json
```diff
@@ -4429,5 +4429,6 @@
       "varchar": "string"
     }
   },
-  "mutationsVersion": "v2"
+  "mutationsVersion": "v2",
+  "mutationsPrefix": ""
```
danieljharvey, line 4433: Before we added this, all the Postgres tests passed as before, after this, we had to go fix lots of things. Which means it works, which is good.

crates/configuration/src/version5/mod.rs
```diff
@@ -51,6 +51,9 @@ pub struct ParsedConfiguration {
     /// Which version of the generated mutation procedures to include in the schema response
     #[serde(default)]
     pub mutations_version: Option<metadata::mutations::MutationsVersion>,
+    /// Provide a custom prefix for generated mutation names. Defaults to mutations version.
```
danieljharvey, line 54: The new config. `serde(default)` means old config that does not include this will get `None`, so we make sure that keeps old behaviour intact.

codedmart (approved):

PR #655, opened 2024-12-03
Fix invalid joins when the order by relationship predicate is null

daniel-chambers:
### What
If a relationship used in an order by specified a null predicate, the SQL generator would fail to add the join conditions when joining the related table into the query. This would result in an incorrect cross product between the table and the related table.

### How

If the relationship predicate is null, we now correctly add the table's join conditions.

Existing tests that assumed all predicates would contain an always true expression have been changed to contain a null predicate and an extra test has been added to cover receiving an always true predicate (to make sure that still works).

Fixes ENG-1386

### DIFF
--- a/crates/query-engine/translation/src/translation/query/sorting.rs
+++ b/crates/query-engine/translation/src/translation/query/sorting.rs
@@ -776,31 +776,32 @@ fn select_for_path_element(
     select.select_list = select_list;
     select.from = Some(from_clause);
 
-    match predicate {
-        None => Ok(select),
+    let predicate_tables = RootAndCurrentTables {
+        root_table: root_and_current_tables.root_table.clone(),
+        current_table: join_table,
+    };
+
+    // generate a condition for this join.
+    let join_condition = relationships::translate_column_mapping(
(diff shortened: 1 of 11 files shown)


daniel-chambers:
As far as I can tell, the `Schema Definitions for Metadata / Metadata schema definitions` build is useless and runs zero tests and has for a long time. Should it just be deleted? I suspect a newer version of nextest is now failing the build if zero tests are detected: https://nexte.st/changelog/#0.9.85

crates/query-engine/translation/src/translation/query/sorting.rs
```diff
@@ -776,31 +776,32 @@ fn select_for_path_element(
     select.select_list = select_list;
     select.from = Some(from_clause);
 
-    match predicate {
-        None => Ok(select),
+    let predicate_tables = RootAndCurrentTables {
```
danieljharvey, line 779: OK, so tl;dr, we were short circuiting too early on `None`.

danieljharvey (approved):

PR #656, opened 2024-12-03
guarantee table names are unique

BenoitRanque:
### What

Table names may conflict with scalar type names.

### How

 Prevent conflicts by:

-- assigning unique aliases to tables during introspection, if needed
-- preserving customized aliases during update operations


crates/configuration/src/version3/mod.rs
```diff
@@ -182,6 +182,16 @@ pub async fn introspect(
     let relevant_type_representations =
         filter_type_representations(&scalar_types, type_representations);
 
+    // build a list of names to ensure they are unique. We assume scalar type names + composite types are already a unique set.
+    let mut type_names: HashSet<TypeName> = scalar_types
+        .iter()
+        .map(|t| t.clone().into_inner())
+        .collect();
```
hallettj, line 189: You could slightly simplify this and avoid cloning

```suggestion
    let mut type_names: HashSet<TypeName> = scalar_types
        .into_iter()
        .map(|t| t.into_inner())
        .collect();
```

crates/configuration/src/version4/mod.rs
```diff
@@ -188,6 +200,85 @@ pub async fn introspect(
...
+}
+
+/// given a collection name and a list of already used type names, get a unique name by adding a suffix if needed
+fn get_unique_collection_name(
+    collection_name: CollectionName,
+    type_names: &HashSet<TypeName>,
+) -> CollectionName {
+    let mut collection_name = collection_name;
+
+    if type_names.contains(&collection_name.clone().into_inner()) {
```
hallettj, line 266: The newtype macro implements `AsRef` for borrowing the inner value so you can do,

```suggestion
    if type_names.contains(collection_name.as_ref()) {
```

crates/configuration/src/version4/mod.rs
```diff
@@ -188,6 +200,85 @@ pub async fn introspect(
...
+/// get collections with names guaranteed unique, preserving customized collection and field names if any
+fn get_aliased_tables(
+    type_names: HashSet<TypeName>,
+    tables: metadata::TablesInfo,
+    old_tables: &metadata::TablesInfo,
+) -> metadata::TablesInfo {
+    let mut type_names = type_names;
+    let mut mapped_tables = BTreeMap::new();
+
+    for (collection_name, table_info) in tables.0 {
```
hallettj, line 213: I think I remember you already verified that this will iterate names in a consistent order so that's good

hallettj (approved):
It might be helpful to have a test here - I don't think you need it to verify correctness, but to avoid future breakage. You have an unwritten rule that type names should be iterated in a consistent order to get consistent aliasing on re-introspection. If someone someday changes the type under `TablesInfo` from `BTreeMap` to `HashMap` then everything will look fine, but there will be a subtle bug of inconsistent aliasing.

Anyway, looks good!

BenoitRanque:
> It might be helpful to have a test here - I don't think you need it to verify correctness, but to avoid future breakage. You have an unwritten rule that type names should be iterated in a consistent order to get consistent aliasing on re-introspection. If someone someday changes the type under `TablesInfo` from `BTreeMap` to `HashMap` then everything will look fine, but there will be a subtle bug of inconsistent aliasing.
> 
> Anyway, looks good!

Good call on the test, but I'm actually quite unsure about how to go about testing for non-deterministic behavior. We already have tests that will tells us if introspection yields a different configuration, and being snapshots these will catch incidents of ordering changing.

But I'm not sure how to specifically write a test for ordering becoming non-deterministic, and have that test try to guarantee failure. Besides running a couple times.

PR #664, opened 2024-12-16
Fix Cockroach Native Query problem

i-am-tom:
### What

Native query tracking fails for CockroachDB because OIDs are 64-bit (despite what the documentation says). However, if we just change the type to `i64`, then the vanilla Postgres native query setup fails because `sqlx` requires each type in Rust to correspond to a specific Postgres type, and thus we can't do some sort of dynamic switching.

### How

We cast absolutely everything to `i64` explicitly in the query. The fact that this works seems to be accidental: because the input type becomes irrelevant, Cockroach determines a different type for `oid` than Postgres, and both are happy. `sqlx` bases its types on information from the database, so it's also happy.

I have tested this with tracking a native query in both Cockroach and Postgres, and both seem to be happy.

### DIFF
--- a/crates/configuration/src/version4/native_operations.rs
+++ b/crates/configuration/src/version4/native_operations.rs
@@ -173,7 +173,7 @@ pub async fn oids_to_typenames(
     for row in rows {
         let schema_name: String = row.schema_name;
         let type_name: String = row.type_name;
-        let oid: i64 = row.oid.into();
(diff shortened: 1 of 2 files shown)


danieljharvey (approved):
This makes sense, so long as the types are the same by the time they reach `sqlx`, everybody is happy.

PR #666, opened 2025-01-01
NDC Spec v0.2.0 support

BenoitRanque:
### What

This PR updates ndc-postgres to ndc spec v0.2.0
This includes a lot of changes to tests. These have been justified in individual commits.


BenoitRanque (commented):
~~Note: Failing tests: we expect failing tests related to the deprecation of the root column comparison.
These will be fixed in a separate PR, to be merged on this one before merging to main.~~

This has now been merged.

crates/configuration/src/version4/to_runtime_configuration.rs
```diff
@@ -73,16 +76,35 @@ fn convert_scalar_types(
...
+                        ),
                     },
                 )
             })
             .collect(),
     )
 }
 
+/// Infer scalar type representation from scalar type name, if necessary. Defaults to JSON representation
+fn convert_or_infer_type_representation(
```
BenoitRanque, line 91: V0.2.0 requires type representation.

Type representation comes from introspection configuration, and may be absent.

So, if type representation is missing, we infer the type based on the name and fetch the corresponding type representation from the default introspection configuration.

crates/tests/tests-common/src/request.rs
```diff
@@ -153,3 +153,42 @@ async fn make_request<Response: for<'a> serde::Deserialize<'a>>(
         )
     })
 }
+
+/// Generate/Update static/query.schema.json
+/// this test should be ignored unless explicitly invoked
+#[ignore]
+#[test]
+fn generate_query_request_schema() {
```
BenoitRanque, line 161: Added these utilities to generate query and mutation request schemas, to validate tests files. These tests are not invoked unless explicitly called upon. There's probably better ways to do this, and we can remove them if there's any feelings against keeping them around.

crates/tests/tests-common/src/router.rs
```diff
@@ -21,9 +21,9 @@ pub async fn create_router(
     )]);
     let setup = PostgresSetup::new(environment);
 
-    let state = ndc_sdk::default_main::init_server_state(setup, &absolute_configuration_directory)
+    let state = ndc_sdk::state::init_server_state(setup, &absolute_configuration_directory)
```
BenoitRanque, line 24: No idea what I did here or why, just made the SDK/compiler happy

Cargo.toml
```diff
@@ -31,9 +31,9 @@ similar_names = "allow"
 too_many_lines = "allow"
 
 [workspace.dependencies]
-ndc-models = { git = "https://github.com/hasura/ndc-spec.git", tag = "v0.1.6" }
-ndc-sdk = { git = "https://github.com/hasura/ndc-sdk-rs.git", tag = "v0.4.0" }
-ndc-test = { git = "https://github.com/hasura/ndc-spec.git", tag = "v0.1.6" }
+ndc-models = { git = "https://github.com/hasura/ndc-spec.git", tag = "v0.2.0-rc.2" }
+ndc-sdk = { git = "https://github.com/hasura/ndc-sdk-rs.git", rev = "643b96b" }
```
danieljharvey, line 35: Can we bump `ndc-sdk` separately before this PR goes in?

Cargo.toml
```diff
@@ -53,15 +53,18 @@ nonempty = "0.10"
 percent-encoding = "2"
 prometheus = "0.13"
 ref-cast = "1"
-reqwest = "0.11"
+reqwest = "0.12"
```
danieljharvey, line 56: Can we bump this separately before the PR goes in? Good not to mix up functional and non-functional changes.

crates/configuration/src/version3/mod.rs
```diff
@@ -1020,6 +1054,16 @@ fn convert_operator_kind(
     match operator_kind {
         metadata::OperatorKind::Equal => query_engine_metadata::metadata::OperatorKind::Equal,
         metadata::OperatorKind::In => query_engine_metadata::metadata::OperatorKind::In,
+        metadata::OperatorKind::LessThan => query_engine_metadata::metadata::OperatorKind::LessThan,
```
danieljharvey, line 1057: Are we sure these are two separate types? I've seen a few confusing re-exports in this repo, they could be pointing at the same thing.

crates/query-engine/translation/tests/goldenfiles/aggregate_distinct_albums/request.json
```diff
@@ -1,4 +1,5 @@
 {
+  "$schema": "../../../../../../static/query.schema.json",
```
danieljharvey, line 2: Generating the schema and putting it in the requests as you have done is a great idea to make tests easier to write, you should do it in a PR before this one and merge it to reduce the amount of noise here.

danieljharvey: Mixing up functional and non-functional changes like this makes the important details easier to miss.

crates/query-engine/translation/src/translation/query/relationships.rs
```diff
@@ -99,6 +99,14 @@ pub fn translate_column_mapping(
         .iter()
         .map(|(source_col, target_col)| {
             let source_column_info = table_info.lookup_column(source_col)?;
+            // target_col contains at least one element, the column being referenced
+            // if nested relationships are supported, it additionally contains the path to navigate to the related column
+            // we currently don't support nested relationships, so erring out if the pattern does not match
+            let [target_col] = target_col.as_slice() else {
```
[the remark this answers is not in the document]
danieljharvey: This would make a very helpful code comment.

crates/configuration/src/version3/mod.rs
```diff
@@ -714,14 +714,48 @@ fn convert_scalar_types(
...
 
+/// Infer scalar type representation from scalar type name, if necessary. Defaults to JSON representation
+fn convert_or_infer_type_representation(
+    representation: Option<query_engine_metadata::metadata::TypeRepresentation>,
+    scalar_type_name: &ScalarTypeName,
+) -> query_engine_metadata::metadata::TypeRepresentation {
+    if let Some(representation) = representation {
+        representation
+    } else {
+        match scalar_type_name.as_str() {
```
danieljharvey, line 736: We already have the `default_base_type_representations` to do this mapping from scalar type name to `TypeRepresentation`.  If this function adds any more mappings we should add them there rather than adding a new layer of indirection. The idea that whatever mappings are in the configuration are what will be returned in the schema is a useful property for understanding all this, and we would need a compelling reason to remove that property.

crates/connectors/ndc-postgres/src/schema/mod.rs
```diff
@@ -30,25 +30,39 @@ pub fn get_schema(
...
-                                    // Thus, we mark all aggregates as having a nullable return
-                                    // type.
-                                    underlying_type: Box::new(models::Type::Named {
-                                        name: function_definition.return_type.clone(),
-                                    }),
+                            match (
+                                function_name.as_str(),
+                                function_definition.return_type.as_str(),
+                            ) {
+                                ("sum", "float8" | "int8") => {
```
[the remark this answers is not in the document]
danieljharvey: All looks good - let's add these comments into the code where it makes sense.

crates/connectors/ndc-postgres/src/schema/mod.rs
```diff
@@ -209,6 +198,42 @@ pub fn get_schema(
                         )
                     })
                     .collect(),
+                    foreign_keys: table
```
danieljharvey, line 201: These just moved in the schema right?

crates/query-engine/metadata/src/metadata/database.rs
```diff
@@ -30,7 +30,7 @@ pub struct ScalarType {
     pub description: Option<String>,
     pub aggregate_functions: BTreeMap<models::AggregateFunctionName, AggregateFunction>,
     pub comparison_operators: BTreeMap<models::ComparisonOperatorName, ComparisonOperator>,
-    pub type_representation: Option<TypeRepresentation>,
+    pub type_representation: TypeRepresentation,
```
danieljharvey, line 33: I am starting to think we should just use `ndc_models::TypeRepresentation` here instead of this type which appears to be a complete copy? Feel bugs will lurk in subtle differences between the two, and I don't understand what the indirection buys us.

crates/query-engine/translation/src/translation/query/aggregates.rs
```diff
@@ -19,38 +19,61 @@ pub fn translate(
...
+                        ),
+                    );
+                    let aggregate_function_call_expression = sql::ast::Expression::FunctionCall {
+                        function: sql::ast::Function::Unknown(function.to_string()),
+                        args: vec![column],
+                    };
+                    // postgres SUM aggregate returns null if no input rows are provided
+                    // however, the ndc spec requires that SUM aggregates over no input rows return 0
+                    // we achieve this with COALESCE, falling back to 0 if the aggregate expression returns null
+                    if function.as_str() == "sum" {
```
danieljharvey, line 65: 👍 Thank you for the explanatory comments.

Cargo.toml
BenoitRanque: That's correct, the only reason we bump it up is to bring in the new models. We want to make sure the `ndc-models` we import directly and the ones from the SDK are the same version.

Cargo.toml
BenoitRanque: No. `ndc-test` needs to be updated alongside `ndc-models` and `ndc-sdk`, and it uses `reqwest` `12`, which is the reason for this change.

crates/query-engine/translation/tests/goldenfiles/aggregate_distinct_albums/request.json
BenoitRanque: Absolutely fair, will split this into it's own PR

danieljharvey: Thanks!

crates/configuration/src/version3/mod.rs
BenoitRanque: These tyes are different: `metadata:OperatorKind` is local to this specific configuration version, and represents a part of a (de)serializable configuration. There's a copy of that for each configuration version.

`query_engine_metadata::metadata::OperatorKind` belongs to the connector runtime configuration. It may change overtime, and a serialized configuration needs to be able to map to it.

That said, in practice these types are equivalent right now, and could be the same type. But I followed what seemed to be the existing pattern of keeping the two separate.

crates/query-engine/translation/tests/goldenfiles/aggregate_limit_offset_order_by/configuration.json
```diff
@@ -1,5 +1,19 @@
 {
-  "version": "4",
+  "version": "5",
```
danieljharvey, line 2: Looks like you're bumping all the test configuration versions too? Perhaps do that in a separate PR too? It's hard to know if any changes in here are meaningful or not.

crates/connectors/ndc-postgres/src/schema/mod.rs
BenoitRanque: Correct!

crates/query-engine/translation/tests/goldenfiles/nested_recursive_relationship/request.json
```diff
@@ -31,15 +32,15 @@
   "collection_relationships": {
     "ArtistAlbums": {
       "column_mapping": {
-        "ArtistId": "ArtistId"
+        "ArtistId": ["ArtistId"]
```
danieljharvey, line 35: This `column_mapping` type change seems to be the main change to request files - is there anything else I should expect to see?

crates/configuration/src/version3/mod.rs
BenoitRanque: v0.2 requires type representation in schema responses, however type representation is optional in the configuration itself.

So this is a piece of backwards compatiblity code, to allow for configurations that are missing type representation.

We could instead create configuration v6, and refuse to run with older configuration versions.

As for the use of `default_base_type_representations`: this mapping only exists for configurations v4 and v5. Here in v3 we replicate that functionality instead.
I prefered copying the code but keeping the different modules separate.

And yes it's a bit cheaty: we have a default when creating a new config, and reuse that default when type representation is missing.

danieljharvey: It's probably best to leave version 3 configs doing whatever they were before, ie falling back to JSON where they have no `TypeRepresentation`, or where they used the deprecated `Number` or `Integer` types.

Also, there is no need for a version 6, it looks like we were already making any mention of `Number` and `Integer` fallback to JSON before this PR, and since we have our own type exposed in the config type, we won't be removing any items and thus making any breaking changes.

crates/query-engine/translation/tests/goldenfiles/aggregate_limit_offset_order_by/configuration.json
BenoitRanque: agreed, will put that with the one adding the schemas

crates/query-engine/translation/src/translation/helpers.rs
```diff
@@ -47,16 +47,68 @@ pub struct NativeQueryInfo {
...
-/// For the root table in the query, and for the current table we are processing,
+/// For the current table we are processing, and all ancestor table up to the closest Query,
 /// We'd like to track what is their reference in the query (the name we can use to address them,
 /// an alias we generate), and what is their name in the metadata (so we can get
 /// their information such as which columns are available for that table).
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct RootAndCurrentTables {
-    /// The root (top-most) table in the query.
-    pub root_table: TableSourceAndReference,
+pub struct TableScope {
```
danieljharvey, line 55: Does this `TableScope` refactor still make sense in `ndc-models` 0.1.0? If so I would really like us to apply it in a separate PR before this one, as it's the source of a lot of changes I can see.

crates/query-engine/translation/tests/goldenfiles/aggregate_distinct_albums/request.json
BenoitRanque: Done! https://github.com/hasura/ndc-postgres/pull/680

crates/query-engine/metadata/src/metadata/database.rs
danieljharvey: This could be done in a separate PR to be fair, before or after this one.

crates/query-engine/translation/src/translation/query/filtering.rs
```diff
@@ -741,54 +732,99 @@ pub fn translate_exists_in_collection(
...
-                .lookup_fields_info(&root_and_current_tables.root_table.source)?
-                .lookup_column(name)?;
-
-            let mut field_path = match field_path {
-                None => VecDeque::new(),
-                Some(field_path) => field_path.iter().collect(),
-            };
-            get_column_scalar_type_name(env, &column.r#type, &mut field_path)
-        }
+        // models::ComparisonTarget::RootCollectionColumn { name, field_path } => {
```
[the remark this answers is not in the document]
danieljharvey: Yeah, delete it please, we have Git if we want to see history.

crates/query-engine/translation/src/translation/query/filtering.rs
```diff
@@ -741,54 +732,99 @@ pub fn translate_exists_in_collection(
...
-                        .lookup_fields_info(&root_and_current_tables.current_table.source)?
-                        .lookup_column(name)?;
+            let column = env
+                .lookup_fields_info(&current_table_scope.current_table().source)?
+                .lookup_column(name)?;
 
-                    get_column_scalar_type_name(env, &column.r#type, &mut field_path)
+            get_column_scalar_type_name(env, &column.r#type, &mut field_path)
+        }
+        ndc_models::ComparisonTarget::Aggregate { path, aggregate } => {
```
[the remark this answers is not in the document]
danieljharvey: If it's not necessary, let's remove and do this in a follow-up PR where we can look at it properly.

crates/query-engine/translation/src/translation/helpers.rs
danieljharvey: If we could split out the `TableScope` changes from the `ndc-models` changes then I think this will become a much clearer atomic change.

crates/query-engine/translation/src/translation/query/sorting.rs
```diff
@@ -97,11 +97,15 @@ struct Column(models::FieldName);
 /// An aggregate operation to select from a table used in an order by.
 #[derive(Debug)]
 enum Aggregate {
-    CountStarAggregate,
-    SingleColumnAggregate {
+    StarCount,
```
danieljharvey, line 100: These new names are better, but again, this change is arbitrary and just more noise, this kind of thing should just be a tiny gardening PR.

crates/query-engine/translation/src/translation/query/sorting.rs
```diff
@@ -459,9 +478,23 @@ fn build_select_and_joins_for_order_by_group(
...
-                                        function: function.clone(),
-                                        args: vec![column],
+                                    Some(aggregate) => match aggregate {
+                                        OrderByAggregate::Function { function } => {
+                                            sql::ast::Expression::FunctionCall {
+                                                function: function.clone(),
+                                                args: vec![column],
+                                            }
+                                        }
+                                        OrderByAggregate::CountStar | OrderByAggregate::Count => {
```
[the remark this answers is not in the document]
danieljharvey: I've looked through [here](https://hasura.github.io/ndc-spec/specification/queries/aggregates.html#aggregates) and I'm not sure, perhaps a question for @daniel-chambers ?

crates/query-engine/translation/tests/snapshots/tests__aggregate_limit_offset_order_by.snap
```diff
@@ -23,7 +24,7 @@ FROM
               COUNT("%2_Invoice"."InvoiceId") AS "InvoiceId_count",
               min("%2_Invoice"."Total") AS "Total__min",
               max("%2_Invoice"."Total") AS "Total__max",
-              sum("%2_Invoice"."Total") AS "Total__sum",
+              coalesce(sum("%2_Invoice"."Total"), 0) AS "Total__sum",
```
danieljharvey, line 27: 👍

crates/tests/tests-common/goldenfiles/select_where_in_column.json
```diff
@@ -19,11 +19,8 @@
       "operator": "_in",
       "value": {
         "type": "column",
```
danieljharvey, line 21: What is going on with this change?

crates/query-engine/translation/src/translation/query/sorting.rs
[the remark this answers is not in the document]
daniel-chambers: `CountStar` and `Count(column)` are not semantically the same. As mentioned in that doc, `CountStar` counts all rows, `Count(column)` counts all rows that have a non-null value in `column`. This is actually consistent with the usual behaviour `COUNT(*)` and `COUNT(column)` in SQL. 

I'd need to see the SQL that this generates to know what's actually going on here... because this is counting across a join, you can't use `count(*)` as Benoit pointed out, but I'd like to know what we're actually counting then. I tried to follow the code and it exceeded my 10:30pm patience 😂. My gut says that in this situation a "count(*)" should be count("the join key column"), which may be what's going on here.

Regardless, it needs to satisfy the semantics I mentioned above.

crates/configuration/src/version3/mod.rs
BenoitRanque: Defaulting to JSON is a behavior change. The previous behavior was that since representation was optional in both schema response and configuration, we just returned whatever was in configuration.

This applies to applies to configurations 3,4,5

Confirm we'd rather default to JSON if representation missing, without first trying to infer based on type name?

crates/query-engine/translation/src/translation/helpers.rs
BenoitRanque: TableScope is for v0.2, as it enables supporting named scopes. We can apply it in a subsequent PR, but some tests won't pass until then

crates/tests/tests-common/goldenfiles/select_where_in_column.json
BenoitRanque: > ComparisonValue::Column no longer uses ComparisonTarget to pick the column. Instead, the necessary column and pathing details are inlined onto the enum variant.

This got flattened

crates/configuration/src/version3/mod.rs
danieljharvey: Yes, default to JSON, this is what is assumed if no type representation is provided, so semantically it's no change, it just means the connector chooses JSON instead of engine doing it.

crates/query-engine/translation/src/translation/query/filtering.rs
```diff
@@ -434,18 +445,17 @@ fn translate_comparison_pathelements(
 /// translate a comparison target.
 fn translate_comparison_target(
     env: &Env,
-    state: &mut State,
-    root_and_current_tables: &RootAndCurrentTables,
+    _state: &mut State,
```
danieljharvey, line 448: Remove unused arguments pls.

crates/query-engine/translation/src/translation/query/sorting.rs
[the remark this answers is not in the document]
danieljharvey: Do we have a test that confirms this works? If not, can we add one please?

danieljharvey:
There's a failing e2e test [here](https://github.com/hasura/v3-e2e-testing/blob/8d58eee0368c643853afa604f39e3dd268ad8e49/crates/postgres/tests/goldenfiles/mutations/insert_parent_and_child.graphql) that seems to have an error returning relationships from a mutation, can we add a test for this here please? (and fix whatever comes up, obv)

crates/query-engine/translation/src/translation/query/filtering.rs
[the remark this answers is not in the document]
danieljharvey: Actually, change of heart, let's write tests for it that confirms it works, or remove for a follow-up.

crates/query-engine/translation/src/translation/query/sorting.rs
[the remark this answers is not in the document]
BenoitRanque: The SQL I shared above is from a test so I think we're covered in terms of SQL generation working as intended

crates/connectors/ndc-postgres/src/schema/mod.rs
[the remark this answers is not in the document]
BenoitRanque: Added a comment. Also changed up the code to fetch the referenced (possibly scalar) type, and match based on that representation.
If we fail to match we assume it's not a scalar, and the functions will be custom.

This should be better than matching on scalar type names.

.github/workflows/e2e-tests.yaml
```diff
@@ -69,9 +69,9 @@ jobs:
             ${{ github.workspace }}/ndc-postgres/result/bin/ndc-postgres
         run: |
           npm install
-          just start-postgres-dependencies start-apollo-subgraph
+          just start-postgres_v0_2-dependencies start-apollo-subgraph
           trap 'just stop-everything' EXIT
-          RUST_LOG=DEBUG ./crates/postgres/static/run-postgres-tests.sh '.*' "$NDC_POSTGRES_BINARY" "$ENGINE_BINARY"
+          RUST_LOG=DEBUG ./crates/postgres_v0_2/static/run-postgres-tests.sh '.*' "$NDC_POSTGRES_BINARY" "$ENGINE_BINARY"
```
danieljharvey, line 74: This now explicitly runs the new e2e tests that use `ndc-models 0.2.0`

BenoitRanque:
Merged main into this branch

PR #669, opened 2025-01-09
Update multitenant version in update script

danieljharvey:
### What

We automatically open `ndc-postgres-multitenant` PRs, let's make sure we update the right part of the `Cargo.toml` file


danieljharvey:
Merge this as soon as https://github.com/hasura/ndc-postgres-multitenant/pull/488 is in.

PR #680, opened 2025-01-23
Update test related files

BenoitRanque:
### What

- add json schema files for configuration, queries and mutations
- reference these schemas in configuration and request files
- update configuration files to version 5
- add manually triggered tests to generate the schema files


danieljharvey (approved):
Amazing, thanks for splitting this out.

PR #685, opened 2025-01-30
Fixed inconsistent handling of missing vs null in mutations and native operations

daniel-chambers:
### What
This PR fixes inconsistent handling of missing arguments/object properties in native operations and v2 mutations. In particular:
- Native operations will now interpret missing arguments as null values for that argument, instead of causing an error.
- Pre and post-check arguments in v2 mutations can now either be missing or null and both will be interpreted as an always true predicate. Previously a null value would have caused an error.
- In v2 update mutations update columns explicitly set to null (as opposed to being missing or being set with their `_set` value object) are now correctly interpreted as "no update should be made to that column", instead of causing an error.

### How
When reading arguments for native operations in `crates/query-engine/translation/src/translation/query/native_queries.rs`, we check if the argument type is nullable and if it is and the argument is missing, we assume a null value for the argument. If the argument type is not nullable, then we return an error as before, since we can't assume null.

Pre and post-check arguments are now deserialized using the new `get_nullable_predicate_argument` function in `crates/query-engine/translation/src/translation/mutation/v2/common.rs`. This function allows for both missing and null properties and in both cases coalesces them into the always true predicate.

For v2 update mutations update columns, we now handle the case where the column property is null and treat it as a no-op (see `parse_update_columns`)

### DIFF
--- a/crates/tests/tests-common/goldenfiles/mutations/v2_update_defaults.json
+++ b/crates/tests/tests-common/goldenfiles/mutations/v2_update_defaults.json
@@ -62,7 +62,8 @@
         "key_id": 3,
         "update_columns": {
           "name": { "_set": "Winston" },
-          "height_cm": { "_set": 700 }
+          "height_cm": { "_set": 700 },
+          "height_in": null
         },
         "pre_check": {
           "type": "unary_comparison_operator",
--- a/crates/tests/tests-common/goldenfiles/mutations/v2_update_no_checks.json
+++ b/crates/tests/tests-common/goldenfiles/mutations/v2_update_no_checks.json
@@ -3,11 +3,14 @@
   "operations": [
     {
       "type": "procedure",
-      "name": "delete_PlaylistTrack_by_PlaylistId_and_TrackId",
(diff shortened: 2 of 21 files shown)


crates/tests/tests-common/goldenfiles/mutations/v2_update_defaults.json
```diff
@@ -62,7 +62,8 @@
         "key_id": 3,
         "update_columns": {
           "name": { "_set": "Winston" },
-          "height_cm": { "_set": 700 }
+          "height_cm": { "_set": 700 },
+          "height_in": null
```
daniel-chambers, line 66: This checks null handling of update columns properties.

crates/tests/tests-common/goldenfiles/mutations/v2_update_no_checks.json
daniel-chambers, line 1: This checks missing and null handling of pre and post checks in updates.

BenoitRanque (approved):
This all looks good, thanks for the very generous comments explaining what's going on

PR #687, opened 2025-01-31
add native toolchain definition to connector metadata

paritosh-08:
### What

This PR adds the native toolchain definition to the connector metadata.

It also adds two commands in the initial metadata generated by the CLI:
```yaml
nativeToolchainDefinition:
  commands:
    start:
      type: ShellScript
      bash: |-
        #!/usr/bin/env bash
        set -eu -o pipefail
        HASURA_CONFIGURATION_DIRECTORY="$HASURA_PLUGIN_CONNECTOR_CONTEXT_PATH" "$HASURA_DDN_NATIVE_CONNECTOR_DIR/ndc-postgres" serve
      powershell: |-
        $ErrorActionPreference = "Stop"
        $env:HASURA_CONFIGURATION_DIRECTORY="$env:HASURA_PLUGIN_CONNECTOR_CONTEXT_PATH"; & "$env:HASURA_DDN_NATIVE_CONNECTOR_DIR\ndc-postgres.exe" serve
    update:
      type: ShellScript
      bash: |-
        #!/usr/bin/env bash
        set -eu -o pipefail
        "$HASURA_DDN_NATIVE_CONNECTOR_PLUGIN_DIR/hasura-ndc-postgres" update
      powershell: |-
        $ErrorActionPreference = "Stop"
        & "$env:HASURA_DDN_NATIVE_CONNECTOR_PLUGIN_DIR\hasura-ndc-postgres.exe" update
```

<details>
<summary>Click here for the CLI command output</summary>

```
❯ ndc-postgres-cli initialize --with-metadata
❯ cat .hasura-connector/connector-metadata.yaml
packagingDefinition:
  type: PrebuiltDockerImage
  dockerImage: ghcr.io/hasura/ndc-postgres:343b8b3
supportedEnvironmentVariables:
- name: CONNECTION_URI
  description: The PostgreSQL connection URI
  defaultValue: postgresql://read_only_user:readonlyuser@35.236.11.122:5432/v3-docs-sample-app
- name: CLIENT_CERT
  description: The SSL client certificate (Optional)
  defaultValue: ''
- name: CLIENT_KEY
  description: The SSL client key (Optional)
  defaultValue: ''
- name: ROOT_CERT
  description: The SSL root certificate (Optional)
  defaultValue: ''
commands:
  update: hasura-ndc-postgres update
cliPlugin:
  name: ndc-postgres
  version: 343b8b3
dockerComposeWatch:
- path: ./
  target: /etc/connector
  action: sync+restart
nativeToolchainDefinition:
  commands:
    start:
      type: ShellScript
      bash: |-
        #!/usr/bin/env bash
        set -eu -o pipefail
        HASURA_CONFIGURATION_DIRECTORY="$HASURA_PLUGIN_CONNECTOR_CONTEXT_PATH" "$HASURA_DDN_NATIVE_CONNECTOR_DIR/ndc-postgres" serve
      powershell: |-
        $ErrorActionPreference = "Stop"
        $env:HASURA_CONFIGURATION_DIRECTORY="$env:HASURA_PLUGIN_CONNECTOR_CONTEXT_PATH"; & "$env:HASURA_DDN_NATIVE_CONNECTOR_DIR\ndc-postgres.exe" serve
    update:
      type: ShellScript
      bash: |-
        #!/usr/bin/env bash
        set -eu -o pipefail
        "$HASURA_DDN_NATIVE_CONNECTOR_PLUGIN_DIR/hasura-ndc-postgres" update
      powershell: |-
        $ErrorActionPreference = "Stop"
        & "$env:HASURA_DDN_NATIVE_CONNECTOR_PLUGIN_DIR\hasura-ndc-postgres.exe" update
```

</details>

### DIFF
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -138,12 +138,12 @@ async fn initialize(with_metadata: bool, context: Context<impl Environment>) -> 
             native_toolchain_definition: Some(NativeToolchainDefinition {
                 commands: vec![
                     ("start".to_string(), metadata::CommandDefinition::ShellScript {
-                        bash: "#!/usr/bin/env bash\nset -eu -o pipefail\nHASURA_CONFIGURATION_DIRECTORY=\"$HASURA_PLUGIN_CONNECTOR_CONTEXT_PATH\" ndc-postgres serve".to_string(),
-                        powershell: "$ErrorActionPreference = \"Stop\"\n$env:HASURA_CONFIGURATION_DIRECTORY=\"$env:HASURA_PLUGIN_CONNECTOR_CONTEXT_PATH\"; & ndc-postgres.exe serve".to_string(),
+                        bash: "#!/usr/bin/env bash\nset -eu -o pipefail\nHASURA_CONFIGURATION_DIRECTORY=\"$HASURA_PLUGIN_CONNECTOR_CONTEXT_PATH\" \"$HASURA_DDN_NATIVE_CONNECTOR_DIR/ndc-postgres\" serve".to_string(),
+                        powershell: "$ErrorActionPreference = \"Stop\"\n$env:HASURA_CONFIGURATION_DIRECTORY=\"$env:HASURA_PLUGIN_CONNECTOR_CONTEXT_PATH\"; & \"$env:HASURA_DDN_NATIVE_CONNECTOR_DIR\\ndc-postgres.exe\" serve".to_string(),
                     }),
                     ("update".to_string(), metadata::CommandDefinition::ShellScript {
-                        bash: "#!/usr/bin/env bash\nset -eu -o pipefail\n\"$HOME/.ddn/plugins/store/ndc-postgres/$POSTGRES_VERSION/hasura-ndc-postgres\" update".to_string(),
-                        powershell: "$ErrorActionPreference = \"Stop\"\n& \"$env:USERPROFILE\\.ddn\\plugins\\store\\ndc-postgres\\$env:POSTGRES_VERSION\\hasura-ndc-postgres.exe\" update".to_string(),
+                        bash: "#!/usr/bin/env bash\nset -eu -o pipefail\n\"$HASURA_DDN_NATIVE_CONNECTOR_PLUGIN_DIR/hasura-ndc-postgres\" update".to_string(),
+                        powershell: "$ErrorActionPreference = \"Stop\"\n& \"$env:HASURA_DDN_NATIVE_CONNECTOR_PLUGIN_DIR\\hasura-ndc-postgres.exe\" update".to_string(),
                     }),
                 ].into_iter().collect(),
             })
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -9,6 +9,7 @@ mod native_operations;
 use std::path::PathBuf;
 
 use clap::Subcommand;
+use metadata::NativeToolchainDefinition;
 use tokio::fs;
 
 use ndc_postgres_configuration as configuration;
@@ -134,6 +135,20 @@ async fn initialize(with_metadata: bool, context: Context<impl Environment>) -> 
                 action: metadata::DockerComposeWatchAction::SyncAndRestart,
                 ignore: vec![],
             }],
+            native_toolchain_definition: Some(NativeToolchainDefinition {
+                commands: vec![
+                    ("start".to_string(), metadata::CommandDefinition {
+                        command_type: metadata::CommandType::ShellScript,
+                        bash: "#!/usr/bin/env bash\nset -eu -o pipefail\nHASURA_CONFIGURATION_DIRECTORY=\"$HASURA_PLUGIN_CONNECTOR_CONTEXT_PATH\" ndc-postgres serve".to_string(),
+                        powershell: String::new(),
+                    }),
+                    ("update".to_string(), metadata::CommandDefinition {
+                        command_type: metadata::CommandType::ShellScript,
+                        bash: "#!/usr/bin/env bash\nset -eu -o pipefail\n\"$HOME/.ddn/plugins/bin/hasura-ndc_postgres\" update".to_string(),
+                        powershell: String::new(),
+                    }),
+                ].into_iter().collect(),
+            })
         };
 
         fs::write(metadata_file, serde_yaml::to_string(&metadata)?).await?;
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -137,13 +137,12 @@ async fn initialize(with_metadata: bool, context: Context<impl Environment>) -> 
             }],
             native_toolchain_definition: Some(NativeToolchainDefinition {
                 commands: vec![
-                    ("start".to_string(), metadata::CommandDefinition {
-                        command_type: metadata::CommandType::ShellScript,
+                    ("start".to_string(), metadata::CommandDefinition::ShellScript {
                         bash: "#!/usr/bin/env bash\nset -eu -o pipefail\nHASURA_CONFIGURATION_DIRECTORY=\"$HASURA_PLUGIN_CONNECTOR_CONTEXT_PATH\" ndc-postgres serve".to_string(),
                         powershell: String::new(),
+
                     }),
-                    ("update".to_string(), metadata::CommandDefinition {
-                        command_type: metadata::CommandType::ShellScript,
+                    ("update".to_string(), metadata::CommandDefinition::ShellScript {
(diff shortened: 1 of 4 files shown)


crates/cli/src/lib.rs
```diff
@@ -134,6 +135,20 @@ async fn initialize(with_metadata: bool, context: Context<impl Environment>) ->
                 action: metadata::DockerComposeWatchAction::SyncAndRestart,
                 ignore: vec![],
             }],
+            native_toolchain_definition: Some(NativeToolchainDefinition {
+                commands: vec![
+                    ("start".to_string(), metadata::CommandDefinition {
+                        command_type: metadata::CommandType::ShellScript,
+                        bash: "#!/usr/bin/env bash\nset -eu -o pipefail\nHASURA_CONFIGURATION_DIRECTORY=\"$HASURA_PLUGIN_CONNECTOR_CONTEXT_PATH\" ndc-postgres serve".to_string(),
```
daniel-chambers, line 142: Are you sure this will work? This is not what I expect to see passed to `bash -c` as per the specification for native toolchain commands. This looks like what should be in a script file inside the connector definition's .hasura-connector folder. That's what's done in the nodejs lambda connector.

crates/cli/src/lib.rs
```diff
@@ -134,6 +135,20 @@ async fn initialize(with_metadata: bool, context: Context<impl Environment>) ->
...
+                        powershell: String::new(),
```
daniel-chambers, line 143: Please implement the same thing in Powershell. We have Windows customers.

crates/cli/src/lib.rs
SandeepSamba: Either of them works. The `bash` field can point to a script file or contain the script itself.

daniel-chambers: @SandeepSamba Interesting; that's not in the spec afaik. Do you detect this and run it differently? (ie not "bash -c")

SandeepSamba: No, we just dump `bash -c "whatever that is there in the field"` and it works

daniel-chambers: Hrm. When I try this manually in my terminal it doesn't work, which is why I was questioning it.

For example,

```
bash -c '#!/usr/bin/env bash\nset -eu -o pipefail\necho really\necho test > test.txt'
```

neither echo "really" nor writes "test" into `test.txt`.

daniel-chambers: Oh okay, this is just my bad. If I use single quotes, it's not substituting the `\n`s. If I do this:

```
bash -c '#!/usr/bin/env bash
set -eu -o pipefail
echo really
echo test > test.txt'
```

It works. OK all good then!

SandeepSamba: Now this is super weird, same result when I tried the command you tried, but if I put the the following in a file `test.sh` and do `bash -c "$(cat ./test.sh)"`, it works!!!
```bash
#!/usr/bin/env bash
set -eu -o pipefail
echo really
echo test > test.txt
```

daniel-chambers: That doesn't surprise me now. My mistake was using single quotes and then using escape sequences. :)

SandeepSamba: Great minds think alike :)

crates/cli/src/lib.rs
SandeepSamba: The native runtime for postgres is targeted to be run inside the [Native DDN workspace](https://github.com/hasura/ddn-workspace/blob/main/native/Dockerfile) only and this script assumes the binary is present in so and so location, so this is not a general script that all users in UNIX systems can run (Postgres is setup in workspace like [this](https://github.com/hasura/ddn-workspace/blob/main/native/Dockerfile#L77-L84)). Also, I don't think there is a windows binary for the Postgres Connector today (I mean the connector itself, not the plugin). So, lets skip the powershell script for now?

SandeepSamba: Makes sense, Added the powershell config [here](https://github.com/hasura/ddn-workspace/blob/main/native/sample-projects/sample-postgres/app/connector/mypostgres/.hasura-connector/connector-metadata.yaml#L4-L23). @paritosh-08 please update using this snippet

crates/cli/tests/snapshots/initialize_tests__initialize_directory_with_metadata.snap
```diff
@@ -27,3 +27,23 @@ dockerComposeWatch:
...
+        $env:HASURA_CONFIGURATION_DIRECTORY="$env:HASURA_PLUGIN_CONNECTOR_CONTEXT_PATH"; & ndc-postgres.exe serve
+    update:
+      type: ShellScript
+      bash: |-
+        #!/usr/bin/env bash
+        set -eu -o pipefail
+        "$HOME/.ddn/plugins/store/ndc-postgres/$POSTGRES_VERSION/hasura-ndc-postgres" update
+      powershell: |-
+        $ErrorActionPreference = "Stop"
+        & "$env:USERPROFILE\.ddn\plugins\store\ndc-postgres\$env:POSTGRES_VERSION\hasura-ndc-postgres.exe" update
```
daniel-chambers, line 49: This is dodgy AF. We need a better way of locating the CLI plugin binary. Shouldn't it just be on the PATH like for regular CLI plugins?

We must avoid hardcoding paths like this. This gets baked into specific immutable versions of the connector and if anything needs to change about how we install these in the future, the versions of the connector with this hardcoded in it will be broken. If we need to add something to a spec to fix this, we should, before we back ourselves into a corner.

daniel-chambers: Where does that POSTGRES_VERSION env var even come from? The CLI needs to set some env var or configure the PATH so that the correct version is just automatically located. How does POSTGRES_VERSION even work for other non-Postgres connectors?

SandeepSamba: The plugins are not on path for normal case as well. They are present in this exact directory and CLI invokes then by `cd`-ing into this directory. The connector and plugin is not expected to be run using the native runtime configuration outside the [DDN native workspace](https://github.com/hasura/ddn-workspace/tree/main/native). If the users want to do that, they can either mimic this config or edit the file manually to suit their setup. 

The version env vars are set in the workspace [image](https://github.com/hasura/ddn-workspace/blob/main/native/Dockerfile#L165-L176) and sourced from [here](https://github.com/hasura/ddn-workspace/blob/main/native/Dockerfile#L2-L7). Each connector has its own env var and the scripts are different for each of them (eg for [oracle](https://github.com/hasura/ddn-workspace/blob/main/native/sample-projects/sample-oracle/app/connector/myoracle/.hasura-connector/connector-metadata.yaml#L4-L36))

I do agree that we have hijacked the native runtime configuration to make it work with the native workspace instead of a generic native runtime, but as of today native runtime configuration is not at all used. Even this codepath of using this config is behind an env var toggle [here](https://github.com/hasura/ddn-workspace/blob/main/native/Dockerfile#L162). Only if this env var is true, CLI eveer uses the native runtime config.

daniel-chambers: So if the CLI sets the working directory to where the CLI plugins are before invoking the shell command here, then we don't need all this hardcoded pathing, right?

My concern is that the way it is done here is extremely highly coupled to a very specific implementation. There shouldn't be weird implicit knowledge like "oh this uses some connector-custom env var that we hardcode in the workspace docker image", or "these files are guaranteed to sit in this very specific location on the disk".  We want to decouple the implementation details of the workspace from how the connector consumes this. If the CLI sets the working directory (ie `cd`) to the plugin directory before invoking the shell command, then the command can simply be this:

```
#!/usr/bin/env bash
set -eu -o pipefail
hasura-ndc-postgres update
```

SandeepSamba: But by spec, When native runtime scripts are called, the `cwd` [should](https://github.com/hasura/ndc-hub/blob/main/rfcs/0004-native-packaging.md#proposed-solution) be the `.hasura-connector` directory. So we cannot do the `cd`-ing when native runtime script is called. During a normal binary plugin invocation, CLI does this automatically

SandeepSamba: Resolved with https://github.com/hasura/ndc-hub/blob/main/rfcs/0012-native-workspace-integration.md

daniel-chambers (approved):
This looks great now, thanks for working through the script env var issues. 🙂

PR #688, opened 2025-02-03
Remove mold suggestion in `.envrc.example`

danieljharvey:
### What

It's totally reasonable that a user would just copy `.envrc.example` to `.envrc` and then get going, however this means we try and use `mold`, which we do not install in the Nix flake, and does not work on MacOS. Let's make the default path as conservative and likely to work as possible.

### How

Remove `mold` advice from `.envrc.example`, make `CARGO_INCREMENTAL` flag opt in.

### DIFF
--- a/.envrc.example
+++ b/.envrc.example
@@ -7,7 +7,4 @@
 use flake
 
 # enable incremental compilation
-export CARGO_INCREMENTAL='true'
-
-# use `mold` as your linker, for faster linking (only on Linux)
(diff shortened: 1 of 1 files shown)


BenoitRanque (approved):
Are we sure we should not keep cargo incremental as default?

danieljharvey:
> Are we sure we should not keep cargo incremental as default?

I'd rather give people a menu of things to explicitly uncomment, but don't feel strongly about it by any means.

PR #713, opened 2025-03-05
Add subquery to separate where clause from lateral joins used for relationships

BenoitRanque:
### What

This PR copies some patterns used in v2, which it turns out yield better performance on cockroachdb.

### How

We add an additional subquery wherever we directly reference a source table, to make sure the where clause for that table is applied before any lateral joins.

### Ordering

Ordering and limit are also moved into the subquery.

V2 instead does the following:
- if ordering across a relationship, put the join and order by clause outside the subquery, at the same level as other relationships. also put any limit clause there
- if ordering on a local field, put the order by and limit inside the subquery. additionally, also put a copy of the order by clause outside the subquery, at the same level as relationships

We do not replicate this behavior, and the last bit worries me.
Will ordering be preserved without the second order by clause?

Commiting as is for now but will probably add the clause.

### DIFF
--- a/crates/query-engine/translation/tests/snapshots/tests__sorting_by_nested_relationship_count.snap
+++ b/crates/query-engine/translation/tests/snapshots/tests__sorting_by_nested_relationship_count.snap
@@ -3,7 +3,7 @@ source: crates/query-engine/translation/tests/tests.rs
 expression: result
 ---
 SELECT
-  coalesce(json_agg(row_to_json("%3_universe")), '[]') AS "universe"
+  coalesce(json_agg(row_to_json("%4_universe")), '[]') AS "universe"
 FROM
   (
     SELECT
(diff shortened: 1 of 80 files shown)


BenoitRanque (commented):
Two things of note:
- we consume one more ID due to the additional subquery. an unfortunate side effect is that all queries are affected, even ones that only have aggregates. Expect most otherwise unaffected SQL generation tests to see a bump by 1 in the ids
- we currently have failing tests due to changes in query plans. Fixing those and making sure they're not regressions

danieljharvey:
> Will ordering be preserved without the second order by clause?

Is there a test we can add to validate we still have the correct behaviour?

crates/query-engine/translation/tests/snapshots/tests__sorting_by_nested_relationship_count.snap
```diff
@@ -3,42 +3,47 @@ source: crates/query-engine/translation/tests/tests.rs
...
+                        FROM
+                          "public"."Album" AS "%1_ORDER_PART_Album"
+                        WHERE
+                          (
+                            "%0_Artist"."ArtistId" = "%1_ORDER_PART_Album"."ArtistId"
+                          )
+                      ) AS "%1_ORDER_PART_Album"
+                  ) AS "%2_ORDER_FOR_Artist" ON ('true')
+                ORDER BY
+                  "%2_ORDER_FOR_Artist"."AlbumId" DESC
```
BenoitRanque, line 41: V2 also has a copy of the order by clause on the parent subquery to this one.
I presume there's the potential for ordering to be lost when joining with other tables?

Are we comfortable without adding that additional clause?

danieljharvey: I would feel a lot more comfortable if we added a test so we know we're not creating a regression.

danieljharvey: Looks like we have a test for ordering across relationships already, so this is fine IMO. https://github.com/hasura/ndc-postgres/blob/210ea0510a8621c32e515d017d0a44272f33de78/crates/tests/tests-common/goldenfiles/sorting_by_nested_relationship_column_with_predicate.json#L4

danieljharvey: By all means open a ticket for further investigation, I am satisfied for the moment though.

PR #718, opened 2025-03-07
Bump `ring` version

danieljharvey:
### What

`cargo audit` reports it as unsafe, so bumping dep as advised.


BenoitRanque (approved):

danieljharvey:
Merge keeps getting ruined by errors in the Github action that installs `protoc`.

PR #734, opened 2025-03-24
Remove haskell CI so we can better integrate with rest of CI, which is bash scripts

BenoitRanque:
### What

Remove haskell CI step, so future PRs can better integrate with rest of CI which is bash scripts.

### How

`changelog.hs` would do the following to changelog.md:
1. remove the top 4 lines
```md
# Changelog

## [Unreleased]

```
2. Prepend the text with the following changelog template:
```md
# Changelog

## [Unreleased]

### Added

### Changed

### Fixed

## [<new-version>] - <date>
```
3: Find a line that starts with `[Unreleased]: ` and replace it with a reference to the newly created tag, and update the reference to unreleased code

We replace this with a bash script that does much of the same. Instead of deleting the top 4 lines, we replace the `## [Unreleased]` header with the new template.

This change is a no-op, but will facilitate upcoming changes to CI.


danieljharvey (approved):
❤️

PR #780, opened 2025-06-26
Dynamic connection strings (WIP

BenoitRanque:
this PR implements dynamic connection strings. Still a work in progress

### DIFF
--- a/crates/connectors/ndc-postgres/src/state.rs
+++ b/crates/connectors/ndc-postgres/src/state.rs
@@ -23,7 +23,7 @@ const CONNECTION_STRING_ARGUMENT: &str = "connection_string";
 /// State for our connector.
 #[derive(Debug)]
 pub struct State {
-    pub pool: Pool,
+    pub pool_manager: PoolManager,
     pub query_metrics: metrics::Metrics,
     pub configuration_metrics: ndc_postgres_configuration::Metrics,
 }
@@ -31,14 +31,15 @@ pub struct State {
 type DatabaseConnectionName = String;
 type DatabaseConnectionString = String;
 
+/// A connection pool bundled with database info and a unique index
 #[derive(Debug)]
-pub struct PoolInstance {
+pub struct PoolContext {
     pub pool: PgPool,
     pub database_info: DatabaseInfo,
     pub index: usize,
 }
 
-impl PoolInstance {
+impl PoolContext {
     /// Create a connection pool with default settings.
     /// - <https://docs.rs/sqlx/latest/sqlx/pool/struct.PoolOptions.html>
     async fn new(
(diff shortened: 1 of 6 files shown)


crates/configuration/src/values/redacted.rs
```diff
@@ -0,0 +1,28 @@
+use serde::Deserialize;
```
danieljharvey, line 1: This is a neat thing, you should merge this before any of this other stuff.

crates/connectors/ndc-postgres/src/state.rs
```diff
@@ -19,52 +25,309 @@ use query_engine_execution::metrics;
...
+}
+
+#[derive(Debug)]
+pub enum Pool {
+    Static {
+        pool: Arc<PoolInstance>,
+    },
+    Named {
+        fallback_pool: Option<Arc<PoolInstance>>,
+        pools: Arc<RwLock<BTreeMap<String, Arc<PoolInstance>>>>,
```
danieljharvey, line 94: Are the `String` keys here some sort of named connection identifier? If so, let's make a newtype wrapper around `String` so that's easy to see from the type signatures.

crates/connectors/ndc-postgres/src/state.rs
```diff
@@ -19,52 +25,309 @@ use query_engine_execution::metrics;
...
 /// Create a connection pool and wrap it inside a connector State.
 pub async fn create_state(
-    connection_uri: &str,
+    connection_settings: &ConnectionSettings,
     environment: &impl Environment,
     pool_settings: &PoolSettings,
     metrics_registry: &mut prometheus::Registry,
     version_tag: ndc_postgres_configuration::VersionTag,
 ) -> Result<State, InitializationError> {
-    let connection_url: Url = connection_uri
```
danieljharvey, line 36: This is moved into `PoolInstance` right?

crates/connectors/ndc-postgres/src/state.rs
```diff
@@ -19,52 +25,309 @@ use query_engine_execution::metrics;
...
+                    fallback_pool,
+                    pools: Arc::new(RwLock::new(BTreeMap::new())),
+                    ssl: ssl.clone(),
+                    pool_settings: pool_settings.clone(),
+                })
+            }
+        }
+    }
+
+    pub async fn aquire(
```
danieljharvey, line 177: Spelling nit: 

```suggestion
    pub async fn acquire(
```

crates/configuration/src/version5/connection_settings.rs
```diff
@@ -28,6 +35,27 @@ impl DatabaseConnectionSettings {
...
+    Dynamic {
+        fallback_to_static: bool,
+    },
+}
+
+#[derive(Clone, PartialEq, Eq, Debug, Deserialize, Serialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub enum ConnectionUris {
+    Variable(Variable),
+    Map(BTreeMap<String, ConnectionUri>),
```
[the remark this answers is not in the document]
danieljharvey: Or are these the types we create after we've done that?

crates/connectors/ndc-postgres/src/state.rs
BenoitRanque: Yep, that was the intent and I literally forgot to use the newtype I created for it :)

crates/connectors/ndc-postgres/src/state.rs
BenoitRanque: Appreciate this

PR #817, opened 2025-08-01
Fix stack overflow in OR expression processing and add OR-to-IN optimization

codedmart:
### What
This PR addresses a critical stack overflow issue that occurs when processing large OR expressions in query predicates. The current implementation creates deeply nested left-heavy expression trees that can exhaust the thread stack, causing runtime crashes with errors like:
```
thread 'tokio-runtime-worker' has overflowed its stack
fatal runtime error: stack overflow
```
Additionally, this PR introduces an optimization that converts multiple OR expressions with equality comparisons on the same column into more efficient IN operations.

#### Stack Overflow Fix:

* Replaced the try_fold approach in OR/AND expression processing with balanced binary tree construction
* Added fold_or_balanced() and fold_and_balanced() helper functions that create balanced trees instead of left-heavy chains
* This prevents deep recursion during SQL generation and evaluation

#### OR-to-IN Optimization:

* Added try_optimize_or_to_in() function that detects patterns like:
```
-- Before: 
column = 'value1' OR column = 'value2' OR column = 'value3'

-- After:
column IN ('value1', 'value2', 'value3')
```
* The optimization runs before standard OR processing and only applies when:
  * All expressions are binary equality comparisons (_eq)
  * All comparisons target the same column
  * All values are scalar (not column references)
  * There are at least 2 values to optimize

#### Implementation Details:

* Modified translate_expression_with_joins() in filtering.rs to use balanced tree construction
* Added optimization check at the beginning of OR expression processing
* Leveraged existing IN operator infrastructure for the optimized expressions
* Maintained backward compatibility - non-optimizable OR expressions fall back to the improved balanced tree approach

This change significantly improves performance for queries with large OR predicates while preventing stack overflow crashes entirely.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1880,8 +1880,8 @@ dependencies = [
 
 [[package]]
 name = "ndc-sdk"
-version = "0.8.0"
-source = "git+https://github.com/hasura/ndc-sdk-rs.git?rev=89e1048#89e1048a6545dee0b047930df0598229434e8f91"
+version = "0.8.1"
+source = "git+https://github.com/hasura/ndc-sdk-rs.git?rev=v0.8.1#84dfb1c667de57d095c72d67ad74fdd3ad1fdaa0"
 dependencies = [
  "async-trait",
  "axum",
@@ -1912,8 +1912,8 @@ dependencies = [
 
 [[package]]
 name = "ndc-sdk-core"
-version = "0.8.0"
-source = "git+https://github.com/hasura/ndc-sdk-rs.git?rev=89e1048#89e1048a6545dee0b047930df0598229434e8f91"
+version = "0.8.1"
+source = "git+https://github.com/hasura/ndc-sdk-rs.git?rev=v0.8.1#84dfb1c667de57d095c72d67ad74fdd3ad1fdaa0"
 dependencies = [
  "async-trait",
  "axum",


crates/query-engine/translation/src/translation/query/filtering.rs
```diff
@@ -60,39 +60,41 @@ pub fn translate_expression_with_joins(
...
+            }
+
+            // Create a balanced AND tree to avoid stack overflow
+            let final_expr = sql::helpers::fold_and_balanced(and_expressions);
+            Ok((final_expr, acc_joins))
         }
         models::Expression::Or { expressions } => {
+            // Try to optimize OR expressions to IN operations first
+            if let Some((optimized_expr, optimized_joins)) =
+                try_optimize_or_to_in(env, state, current_table_scope, expressions)? {
```
danieljharvey, line 80: This other optimization seems nice, but quite magic, let's move this to a follow-up PR in case we need to bisect a production problem later.

crates/query-engine/translation/tests/snapshots/tests__select_where_or_to_in_optimization.snap
```diff
@@ -0,0 +1,54 @@
...
+              "%1_Album"."AlbumId" AS "AlbumId"
+            FROM
+              (
+                SELECT
+                  "%0_Album".*
+                FROM
+                  "public"."Album" AS "%0_Album"
+                WHERE
+                  (
+                    (
```
danieljharvey, line 27: Now the optimization doesn't drop the other "or" values, and can even group multiple columns (here we do both `Title` and `ArtistId`)

crates/query-engine/translation/src/translation/query/filtering.rs
```diff
@@ -913,3 +913,171 @@ fn underlying_type_name(typ: database::Type) -> models::TypeName {
...
+                    .cloned()
+                    .collect::<Vec<_>>(),
+            )
+        }
+    }
+
+    Ok(Some(expressions))
+}
+
+fn create_in_expression(
```
danieljharvey, line 1052: Previously Augment did the `models::Expression` -> `sql::ast::Expression` transformation here too, which was messy and confusing. Good idea though, generally.

danieljharvey (approved):
Yeah, I'm happy this is doing good stuff.

